PR #238, opened 2018-02-06
working yargs replacement for commander

rylandg:
Completely swapped out commander js for the far superior `yargs`. I'll list some of the numerous benefits we get from yargs

- unlimited subcommand nesting
- mandatory/optional positional args
- unified command handling
- inherent superfluous argument enforcement
- automatic bash completion
- async natively works
- typescript definitions for commands
- global options list for individual commands
- working `-t` option for `logs`
- complete help and other output control

I've tested this manually using every command(I hope) and through numerous cli spec runs.

My main goal was a port of our existing interfaces + fix commander specific bugs(such as `-t` for logs). Please keep discussion about general CLI text problems in separate issue threads and only discuss the text that was changed in the port here(keeps us agile and not slow on PRs).

For the record I really like `yargs` and anyone who uses it can immediately tell why it's destroying commander lately.

Jenkins build is failing because of an unrelated issue not locating the correct docker image/tag. I'll debug it tomorrow but it's definitely unrelated.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,84 +1,132 @@
-const commander = require('commander');
-// grab the version to keep things consistent
-const { version } = require('./package.json');
+const yargs = require('yargs');
+const moniker = require('moniker');
+
+const logger = require('./lib/logger');
+const { makeNameValid } = require('./lib/nameUtil');
 const { deployHandler, createHandler, invokeHandler,
-  logHandler, loginHandler, removeHandler, unknownHandler } = require('./lib');
+  logsHandler, loginHandler, removeHandler } = require('./lib');
 
-const actionWrapper = function actionWrapper(action, actionOptions) {
-  return async (options) => {
-    await action({
-      options: Object.assign({}, actionOptions, options, { path: commander.path }),
-      args: commander.args,
-      rawArgs: commander.rawArgs,
-      name: commander.args[commander.args.length - 1]._name,
-    });
-  };
+const handleCommand = async function handleCommand(options, specificHandler) {
+  const numArgs = options._.length;
+  if (numArgs > 1) {
+    yargs.showHelp();
+    logger.error('Too many positional args given for command');
+    process.exit(1);
+  }
+
+  if (!options.function) {
+    yargs.showHelp();
+    logger.error('A valid function name is a required argument');
+    if (options._[0] === 'create') {
+      logger.error('The "create" command allows for the --random(r) flag in lieu of a name');
+    }
+    process.exit(1);
+  }
+  await specificHandler(options);
+  process.exit(0);
 };
 
-// again not the cleanest solution but the practical one while
-// still using commander instead of yargs
-const namedActionHandler = function namedActionHandler(action) {
-  return async (funcName, options) => {
-    await action({
-      options: Object.assign({}, { function: funcName }, options, { path: commander.path }),
-      args: commander.args,
-      rawArgs: commander.rawArgs,
-      name: commander.args[commander.args.length - 1]._name,
-    });
-  };
-};
+yargs
+  .option('path', {
+    alias: 'p',
+    describe: 'Use directory dir. "create" will create this directory if needed.',
+    type: 'string',
+    global: true,
+  })
+  .usage(
+`Binaris command line interface
 
-commander
-  .version(version)
-  .description('Binaris command line interface')
-  .option('-p, --path <path>',
-    // eslint-disable-next-line quotes
-    'Use directory dir. "create" will create this directory if needed.');
+Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
+  )
+  .command('create [function] [options]', 'Create a function from template', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 create [function] [options]')
+      .positional('function', {
+        describe: 'Name of the function to generate',
+        type: 'string',
+      })
+      .option('random', {
+        alias: 'r', describe: 'Generate a random name for your function', type: 'boolean',
+      });
+  }, async (argv) => {
+    if (argv.random) {
+      // eslint-disable-next-line no-param-reassign
+      argv.function = makeNameValid(moniker.choose());
+    }
+    await handleCommand(argv, createHandler);
+  })
+  .command('deploy <function> [options]', 'Deploys a function to the cloud', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 deploy <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to deploy',
+        type: 'string',
+      });
+  }, async (argv) => {
+    await handleCommand(argv, deployHandler);
+  })
+  .command('remove <function> [options]', 'Remove a previously deployed function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 remove <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to remove',
+        type: 'string',
+      });
+  }, async (argv) => {
+    await handleCommand(argv, removeHandler);
+  })
+  .command('invoke <function> [options]', 'Invoke a Binaris function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 invoke <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to invoke',
+        type: 'string',
+      })
+      .option('json', {
+        alias: 'j',
+        describe: 'Path to file containing JSON data',
+        type: 'string',
+      })
+      .option('data', {
+        alias: 'd',
+        describe: 'Data to send with invocation',
+        type: 'string',
+      });
+  }, async (argv) => {
+    await handleCommand(argv, invokeHandler);
+  })
+  .command('logs <function> [options]', 'Print the logs of a function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 logs <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to retrieve logs for',
+        type: 'string',
+      })
+      .option('tail', {
+        alias: 't',
+        describe: 'Outputs logs in "tail -f" fashion',
+        type: 'boolean',
+      });
+  }, async (argv) => {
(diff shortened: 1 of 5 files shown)


cli.js
```diff
@@ -1,84 +1,132 @@
...
-      rawArgs: commander.rawArgs,
-      name: commander.args[commander.args.length - 1]._name,
-    });
-  };
-};
+const handleCommand = async function handleCommand(options, specificHandler) {
+  const numArgs = options._.length;
+  if (numArgs > 1) {
+    yargs.showHelp();
+    logger.error('Too many positional args given for command');
```
lavie: This is a user-facing message? Because if so it has to be cleaner.

cli.js
```diff
@@ -1,84 +1,132 @@
...
-    await action({
-      options: Object.assign({}, { function: funcName }, options, { path: commander.path }),
-      args: commander.args,
-      rawArgs: commander.rawArgs,
-      name: commander.args[commander.args.length - 1]._name,
-    });
-  };
+  if (!options.function) {
+    yargs.showHelp();
+    logger.error('A valid function name is a required argument');
```
lavie: `Missing argument: function name.`

cli.js
```diff
@@ -1,84 +1,132 @@
...
+    if (options._[0] === 'create') {
+      logger.error('The "create" command allows for the --random(r) flag in lieu of a name');
```
lavie: I honestly think this is a malfeature that we should get rid of. I just don't see it having any value, and it's one more thing to support.

cli.js
```diff
@@ -1,84 +1,132 @@
...
-commander
-  .command('create')
-  .description('Create a function from template')
-  .option('-f, --function <name>',
-    'Name of the function to generate. If omitted, a name will be chosen at random')
-  .action(actionWrapper(createHandler));
+yargs
+  .option('path', {
+    alias: 'p',
+    describe: 'Use directory dir. "create" will create this directory if needed.',
```
lavie: This message confuses me. I don't understand what will happen with `dir`.

cli.js
```diff
@@ -1,84 +1,132 @@
...
+  .usage(
+`Binaris command line interface
 
-commander
-  .command('deploy <function>')
-  .description('Deploys a function to the cloud')
-  .action(namedActionHandler(deployHandler));
+Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
+  )
+  .command('create [function] [options]', 'Create a function from template', (yargs0) => {
```
lavie: From which template? Can I specify the template? It's unclear.

lavie: Is `yargs0` the convention for a local yargs variable? Kind of weird..

cli.js
```diff
@@ -1,84 +1,132 @@
...
+    yargs0
+      .usage('Usage: $0 create [function] [options]')
+      .positional('function', {
```
lavie: this is nice.

cli.js
```diff
@@ -1,84 +1,132 @@
...
+      })
+      .option('random', {
+        alias: 'r', describe: 'Generate a random name for your function', type: 'boolean',
+      });
+  }, async (argv) => {
+    if (argv.random) {
+      // eslint-disable-next-line no-param-reassign
+      argv.function = makeNameValid(moniker.choose());
+    }
+    await handleCommand(argv, createHandler);
```
lavie: Where do we validate names?

cli.js
```diff
@@ -1,84 +1,132 @@
...
+        describe: 'Name of the function to deploy',
+        type: 'string',
+      });
+  }, async (argv) => {
+    await handleCommand(argv, deployHandler);
+  })
+  .command('remove <function> [options]', 'Remove a previously deployed function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 remove <function> [options]')
+      .positional('function', {
```
lavie: A shame you have to repeat common arguments' definitions. Should have been possible to reuse the `function` argument's definition.

cli.js
```diff
@@ -1,84 +1,132 @@
...
+    await handleCommand(argv, removeHandler);
+  })
+  .command('invoke <function> [options]', 'Invoke a Binaris function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 invoke <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to invoke',
+        type: 'string',
+      })
+      .option('json', {
```
lavie: We're going to need stding support for piping json pretty soon. i.e. `cat file.json | bn invoke foo`

cli.js
```diff
@@ -1,84 +1,132 @@
...
+        type: 'string',
+      });
+  }, async (argv) => {
+    await handleCommand(argv, invokeHandler);
+  })
+  .command('logs <function> [options]', 'Print the logs of a function', (yargs0) => {
+    yargs0
+      .usage('Usage: $0 logs <function> [options]')
+      .positional('function', {
+        describe: 'Name of the function to retrieve logs for',
```
lavie: > Name of the function to retrieve logs for
A bit verbose. It's clear enough if it's just `Function name`. Obviously, from the context, it is clear that this fetches the logs for it...

cli.js
```diff
@@ -1,84 +1,132 @@
...
-  .command('*', null, { noHelp: true })
-  .description('')
-  .action(actionWrapper(unknownHandler));
-
-commander
-  .parse(process.argv);
-
-if (!process.argv.slice(2).length) {
-  commander.outputHelp();
+if (argv._[0] && commands.indexOf(argv._[0]) === -1) {
```
lavie: Highly unreadable to me. Looks like perl. wtf is it even doing?

package.json
```diff
@@ -23,14 +21,16 @@
     "lodash.get": "^4.4.2",
     "moniker": "^0.1.2",
     "mz": "^2.7.0",
+    "pickle": "git+https://github.com:gobinaris/pickle.git#1.0.0",
```
lavie: please publish pickle as a public npm package and consume it accordingly here. `@binaris/pickle` (just switch it to public after the first publish).

test/CLISpec.yml
```diff
@@ -1,35 +1,30 @@
...
-              create [options]             Create a function from template
-              deploy <function>            Deploys a function to the cloud
-              remove <function>            Remove a previously deployed function
-              invoke [options] <function>  Invoke a Binaris function
+              Binaris command line interface
+              
+              Usage: bn <command> [options]
+              
+              Commands:
+                bn create [function] [options]  Create a function from template
```
lavie: *Love* how the first few commands are aligned with text columns. 😁

test/CLISpec.yml
```diff
@@ -1,35 +1,30 @@
...
+                bn deploy <function> [options]  Deploys a function to the cloud
+                bn remove <function> [options]  Remove a previously deployed function
+                bn invoke <function> [options]  Invoke a Binaris function
+                bn logs <function> [options]    Print the logs of a function
+                bn login                        Login to your Binaris account using an API key
+              
+              Options:
+                --version   Show version number  [boolean]
```
lavie: > [boolean]

seems a bit out of place. It's a "flag", not a "boolean parameter".. It's not like anyone will do `--version false`...

test/CLISpec.yml
```diff
@@ -1,35 +1,30 @@
...
+                --path, -p  Use directory dir. "create" will create this directory if needed.  [string]
+                --help, -h  Show help  [boolean]*
 
 - test: Test login (good-path)
   cleanup:
     - bn remove momentousgiants
```
lavie: nit: let's please use a simpler function name here.

cli.js
rylandg: It is, I would love to change it to something more meaningful and succinct. Any suggestions?

cli.js
rylandg: Much better.

cli.js
rylandg: Sure, although as the person using the CLI the most I have to say I use the feature a lot.

cli.js
rylandg: Don't disagree but also this is just a 1:1 port from the old CLI. I'm not sure we want to keep this functionality anyway but if we do we should definitely change this message.

cli.js
rylandg: yargs0 is just the name for the nested yargs context I'm being passed. Vanilla yargs use cases don't rely on nested subcommands(instead just commands with options) and therefore we need to invoke another yargs context inside the outer command's handler. I didn't think it was worth creating names for different yargs objects but I can if you think it's important.

rylandg: There is currently only 1 template. This is also a straight 1:1 port from the old CLI. I agree with your line of reasoning but am not sure exactly what it should say right now.

cli.js
rylandg: Yea the flexibility is miles ahead of the old option.

cli.js
rylandg: Locally using a basic regex [a-z0-9][58]

cli.js
rylandg: So there actually is some crazy stuff you can do in regards to yargs project setup. You can use inheritance and modules for each specific command definition. That way your commands are more full fledged `~class` definitions. I decided to start with a small scope and use a more simple setup. If we see things getting too redundant and complex we can consider the migration.

cli.js
rylandg: Shouldn't be an issue(but should be an _issue_)

cli.js
rylandg: Agreed.

cli.js
rylandg: I agree it's semi-nasty but it's the canonical yargs way of handling this. This is a fall through check in case a command isn't handled by any of the command handlers. `argv._` holds all of the commands in the current yarg pipeline. That means that `argv._[0]` always holds the first command sent by the user. Then we check versus the master list of valid commands and if it isn't present we inform the user.

package.json
rylandg: sure

PR #240, opened 2018-02-13
First fixes from yargs PR

rylandg:
Test via the spec tests and sdk tests. Invoked all valid commands manually without issue.

These are the first set of fixes based on these comments

https://github.com/gobinaris/binaris/pull/238

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -45,7 +45,7 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
     yargs0
       .usage('Usage: $0 create <function> [options]')
       .positional('function', {
-        describe: 'Name of the function to generate',
+        describe: 'Function name',
         type: 'string',
       });
   }, async (argv) => {
@@ -55,7 +55,7 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
     yargs0
       .usage('Usage: $0 deploy <function> [options]')
       .positional('function', {
-        describe: 'Name of the function to deploy',
+        describe: 'Function name',
         type: 'string',
       });
   }, async (argv) => {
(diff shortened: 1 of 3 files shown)


arielshaqed:
Taking this as DDR, but I don't think I'm the original reviewer for most of the comments that this fixes.  @rylandg consider adding as reviewer people who know what it's all about, so we don't go around in circles.

cli.js
```diff
@@ -1,27 +1,30 @@
...
+  process.exit(1);
+};
+
 const handleCommand = async function handleCommand(options, specificHandler) {
-  const numArgs = options._.length;
-  if (numArgs > 1) {
-    yargs.showHelp();
-    logger.error('Too many positional args given for command');
-    process.exit(1);
+  // eslint-disable-next-line no-param-reassign
```
arielshaqed: Why is this needed, I don't see param reassignment here.

cli.js
```diff
@@ -123,10 +119,10 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   .wrap(null);
 
 const commands = yargs.getCommandInstance().getCommands();
-const argv = yargs.argv;
+// first command pushed to command stack(from user input)
```
arielshaqed: What strange wizardry is this?  It looks like you're using `getCommandInstance` (which I couldn't find in the API docs), in order to manually perform behaviour I would expect to get from `yargs`.

rylandg: Yea that wizardry is basically what `commander` feels like all the time. I'm not sure why it's not directly supported.

cli.js
arielshaqed: 🏓 ping?

cli.js
arielshaqed: "Welcome to the future," basically :-(.

cli.js
rylandg: Sorry was busying focusing on other things, I'll remove it in a subsequent patch.

PR #242, opened 2018-02-16
publish via Makefile directive

rylandg:
Just the first bit of logic needed. Not worrying about Jenkins integration yet will be in a second patch.

Tested via the spec cases and manual testing


rylandg:
I'll check and resubmit if it needs a fix.

rylandg:
Ok done

rylandg:
@lavie mind re-reviewing this please

PR #243, opened 2018-02-16
remove invalid linting comment

rylandg:
Tested with spec test. Just a small cleanup from PR.


michaeladda:
@rylandg merging despite not having merge_if_approved label. Please revert if problematic.

PR #248, opened 2018-03-05
test `bn deploy && bn invoke` functionality

rylandg:
Test for the new bach update. Tested with CLI spec tests and manual testing.

https://github.com/binaris/binaris/tree/feature-wait-for-pods

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -74,6 +74,7 @@
       echo '{"name": "unguessable"}' > invoke.json
   cleanup:
     - bn remove extrasmallaunt
+    - bn remove alwaysucceeds
   steps:
     -   in: bn create extrasmallaunt
         out: |-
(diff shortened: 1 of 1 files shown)


test/spec.yml
```diff
@@ -91,6 +92,12 @@
         out: Hello unguessable!
     -   in: bn invoke extrasmallaunt -j invoke.json
         out: Hello unguessable!
+    -   in: bn create alwaysucceeds
```
lavie: I'm not sure what sort of coverage this really adds. e.g. it doesn't really check that the version being invoked is the version that was deployed, right?

rylandg: Well technically since it's deployed and invoked only once and then removed it should be a guarantee. That being said I'll make the output/message unique to avoid confusion.

rylandg:
Updated

PR #250, opened 2018-03-11
update nodejs function template to new signature

michaeladda:
It can be an even simpler example if we drop `req` and rely only on body.
WDYT?

### DIFF
--- a/lib/functionTemplates/nodejs/function.js
+++ b/lib/functionTemplates/nodejs/function.js
@@ -1,5 +1,4 @@
-exports.handler = (event, context, callback) => {
-  const name = event.queryStringParameters.name || event.body.name || 'World';
-  console.log(`Hello ${name}!`);
-  callback(null, { statusCode: 200, body: `Hello ${name}!` });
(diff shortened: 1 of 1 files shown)


lib/functionTemplates/nodejs/function.js
```diff
@@ -1,5 +1,4 @@
-exports.handler = (event, context, callback) => {
-  const name = event.queryStringParameters.name || event.body.name || 'World';
-  console.log(`Hello ${name}!`);
-  callback(null, { statusCode: 200, body: `Hello ${name}!` });
+exports.handler = async (body, req) => {
+  const name = req.query.name || body.name || 'World';
+  return `Hello ${name}!`;
```
arielshaqed: `` `hello, ${name}` `` (https://goo.gl/ABnskW).

rylandg:
I'm just still not sure about the default `async` method. I know it's not the typical `js` `async` use case(as in called from a local environment) but it's still technically incorrect to declare a method async if it doesn't explicitly `await`.

PR #251, opened 2018-03-12
since command for `bn logs`

rylandg:
@lavie I remedied your issues with my WIP commit excluding the bit regarding the description. I did change the description but didn't find an acceptable way to define complex information about the options. For now we have the examples and the simple description(ISO). 

Clearly don't merge until the Jenkins setup is working again and I add the Jenkinsfile back.


lavie:
why did you rebase? this makes it impossible for me to see the most recent changes, which means I have to rereview the entire thing...

rylandg:
@lavie so first off my bad about that. I have a tendency to squash my private branches before PR's which is really stupid (I know WIP). I actually only changed 3 miniscule things.

1. Added `sec` and `min` as you requested
2. Updated tests and examples to use these new abbrvs
3. Changed default help description to something slightly better.

Thanks again.

rylandg:
Was handled separately

PR #252, opened 2018-03-18
Modify logs code to use new API

bergundy:


bergundy:
This was already pushed to master

PR #253, opened 2018-03-21
bump version to 2.0.0

michaeladda:
My bad. Forgot to do this as part of the new signature changes.


michaeladda:
I'll un-publish old version once the new version is GA (it's currently published with a tag)

PR #254, opened 2018-03-26
Force reload of env vars after `bn` actions

rylandg:
Currently once an `sdk` command is `required` it will maintain the env state that existed when it was required. This change allows env vars to update and therefore propagate without `re-requiring` the modules.

Clearly don't approve this until I add the Jenkinsfile back and such.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "description": "Binaris SDK & CLI",
   "main": "index.js",
   "bin": {


Jenkinsfile
```diff
@@ -1,43 +0,0 @@
-node {
```
arielshaqed: Is this part of the PR?

rylandg: I didn't want to possibly retrigger builds that mess up `staging`/`testing` since we still haven't verified that the Jenkins recycle fixed the issue with the Jenkins job.

rylandg: Also I found a bug so glad it didn't go in yet!

PR #260, opened 2018-04-02
Support both startAfter and follow in same logs request

bergundy:
The log server now supports this, and so should the CLI


rylandg:
@bergundy  I talked to @assaf and the conclusion was that this combination of commands doesn't make sense because tail always starts 'now'.

PR #261, opened 2018-04-11
force env vars to reload during runtime

rylandg:


lib/deploy.js
```diff
@@ -8,7 +8,7 @@ const tgzCompress = promisify(compress);
 
 const { getAPIKey } = require('./userConf');
 const { deploy } = require('../sdk');
-const { invokeEndpoint } = require('../sdk/config');
+const { getInvokeEndpoint } = require('../sdk/config');
```
arielshaqed: Once we release this as an SDK, any such change will require a major version release.  Let's decide on a naming convention?

rylandg: I'll respond to both of your comments here.

1. If you notice previously the environment variables were loaded as static variables directly into the relevant SDK files. This meant that even if the ENV's changed during runtime it wouldn't reflect in the `required` code because the ENV's had been cached (so to speak). Now that an actual function is used the ENV's are pulled dynamically.

2. Agree and we had a small discussion on slack about this. Is an issue the best place or will another medium work better?

rylandg:
Ok I fixed a small issue with the patch and bumped the sleep time on the spec so the tests could pass. Merge this if you approve and I don't see it.

PR #262, opened 2018-04-12
add Jenkinsfile back to the CLI

rylandg:
This was removed due to issues during staging->testing migration. It passes now.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,16 +26,6 @@ node {
                    make lint
                 """
             }
-            stage('Run remote tests') {
-                echo 'Deploying and running tests'
-                build(
-                    job: "bn-cli-trigger",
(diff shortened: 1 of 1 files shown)


rylandg:
@michaeladda I've removed the problematic part and update the version. Once the better Jenkins integration is done I'll create a new PR for it.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
michaeladda: just noticed this now, why did you bump the version? there are no code changes in this PR.

rylandg: You're right I'll rebase and remove that change.

rylandg:
@michaeladda removed ver increment.

PR #263, opened 2018-04-12
Support runtime parameter

bergundy:
This allows us to run cpython2, pypy2, and node8.
Previously only node8 was supported


lib/create.js
```diff
@@ -15,22 +20,26 @@ const templateDir = './functionTemplates/nodejs/';
...
   const finalName = functionName;
   validateName(finalName);
+  const template = runtimeToTemplateDir[runtime];
+  if (!template) {
+    throw new Error(`No template for runtime: ${runtime}`);
+  }
   // parse the templated yml and make the necessary modifications
-  const templatePath = path.join(__dirname, templateDir);
+  const templatePath = path.join(__dirname, templateDir, template);
   const binarisConf = await YMLUtil.loadBinarisConf(templatePath);
```
rylandg: This most likely needs to be changed. I believe this will now expect a separate `binaris.yml` in every template directory (js, py etc) which is redundant but acceptable. But then you add the runtime manually back to the conf which suggests that there is a duplicate `binaris.yml` template in every directory. If each runtime had it's own template dir I would suggest hardcoding, but in lieu of that I would say have them all share a root `binaris.yml` template and update accordingly.

lib/create.js
```diff
@@ -15,22 +20,26 @@ const templateDir = './functionTemplates/nodejs/';
...
   const templateName = YMLUtil.getFuncName(binarisConf);
-  const funcConf = YMLUtil.getFuncConf(binarisConf, templateName);
+  const funcConf = { ...YMLUtil.getFuncConf(binarisConf, templateName), runtime };
   // replace the generic function name with the actual name
   YMLUtil.delFuncConf(binarisConf, templateName);
   YMLUtil.addFuncConf(binarisConf, finalName, funcConf);
   // now write out all the files that have been modified
   const file = funcConf.file;
-  await fse.copy(path.join(__dirname, templateDir, file), path.join(functionPath, file));
+  await fse.copy(path.join(__dirname, templateDir, template, file), path.join(functionPath, file));
```
rylandg: just do `join(templatePath, file)`

lib/binarisYML.js
```diff
@@ -60,13 +49,16 @@ const getFuncName = function getFuncName(binarisConf) {
...
+  for (const section of [fileStr, entryStr, runtimeStr]) {
+    if (typeof funcConf[section] !== 'string') {
+      throw new Error(`${binarisConfFile}: field: <${fileStr}> should be a string`);
+    }
   }
-  if (!Object.prototype.hasOwnProperty.call(funcConf, entryStr)) {
-    throw new Error(`${binarisConfFile}: function missing required field: <${entryStr}>`);
+  const fullPath = path.join(funcDirPath, funcConf[fileStr]);
+  const stats = await fs.stat(fullPath);
+  if (!stats.isFile()) {
```
rylandg: Don't know if we need this (generally it's bad practice to check for existence outside of a read/write).

cli.js
```diff
@@ -44,6 +44,14 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   .command('create <function> [options]', 'Create a function from template', (yargs0) => {
     yargs0
       .usage('Usage: $0 create <function> [options]')
+      // TODO: should runtime be positional (required)?
```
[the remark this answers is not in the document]
lavie: I don't see a way to have a sane default for runtime. If it's a python user, genering a js function would be a real annoying pain point. Has to be explicit I'm afraid.

cli.js
```diff
@@ -44,6 +44,14 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+      .option('runtime', {
+        alias: 'r',
+        choices: ['node8', 'cpython2', 'pypy2'],
+        describe: 'Runtime image name',
```
lavie: `image name`? The term `image` does not belong here. it's leaking the implementation (docker image) and the user doesn't care about it. Should be just `runtime`.

cli.js
```diff
@@ -44,6 +44,14 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   .command('create <function> [options]', 'Create a function from template', (yargs0) => {
     yargs0
       .usage('Usage: $0 create <function> [options]')
+      // TODO: should runtime be positional (required)?
+      .option('runtime', {
+        alias: 'r',
+        choices: ['node8', 'cpython2', 'pypy2'],
```
lavie: `cpython2`-> `python2`

Not sure most python users known enough not to be confused by the term `cpython`. Many of them, I suspect, just know that interpreter as the "regular" python. And `python2` is what the other vendors call it as well.

cli.js
```diff
@@ -44,6 +44,14 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        describe: 'Runtime image name',
+        type: 'string',
+        default: 'node8', // Should this be the default (if any)?
```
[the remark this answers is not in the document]
lavie: As I've said above, I disagree about having node as the default. I don't think there can be a sane default here. Any choice will just annoy users of the other languages.

cli.js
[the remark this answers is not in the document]
lavie: position vs. required: a good question. I don't like required named options. But on the other hand, I particularly dislike having *multiple* positional parameters, because it tends to be hard to remember the order, and also unless the values are very self-explanatory, it makes it hard to read.
e.g. `ln -s ... ...` is a terrible terrible CLI. I never remember which is which, because the values you pass give no indication which is which.
however `git push origin master` is fine and dandy. Because `origin` is never the name of a branch and if you read this line you'll never confuse the two.
so `bn create node8 hello`... is it readable enough? probably. but a bit verbose, too.
Maybe stuff like the runtime we should ask the user interactively if he does not specify. I think that's the way serverless framework do things, for example. 
So here's a proposal: `bn create RUNTIME FUNC_NAME`.
If a user just passes `bn create hello` then it's easy to detect that `hello` is not a valid runtime and give an *informative* error. In which case we can query the user interactively for a runtime.
It will look like this:

```
$ bn create hello

Creating skeleton for function "hello".
Please choose the runtime:
* python2
* node8
* pypy
```
If the user just made a mistake in the runtime name, eg.. `bn create python hello` we can, again, give an informative error message.

lib/binarisYML.js
```diff
@@ -60,13 +49,16 @@ const getFuncName = function getFuncName(binarisConf) {
...
+    throw new Error(`Not a file in path: ${fullPath}`);
```
lavie: This error is user-visible? If so, it should be more informative.

lib/create.js
```diff
@@ -15,22 +20,26 @@ const templateDir = './functionTemplates/nodejs/';
...
  */
-const create = async function create(functionName, functionPath) {
+const create = async function create(functionName, functionPath, runtime) {
   // removing the '-' from monikers string is required because they
   // don't allow the glue string to be empty.
   const finalName = functionName;
   validateName(finalName);
+  const template = runtimeToTemplateDir[runtime];
+  if (!template) {
+    throw new Error(`No template for runtime: ${runtime}`);
```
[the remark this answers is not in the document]
lavie: yea, this is more like an assert.

lib/binarisYML.js
michaeladda: Not related to read and write. Use, don't check. Return informational error if it fails.
(the rationale is it can fail even after checking, so the code must handle the error well, and it's better to do it once).

lib/create.js
```diff
@@ -4,7 +4,12 @@ const path = require('path');
 const { validateName } = require('./nameUtil');
 const YMLUtil = require('./binarisYML');
 
-const templateDir = './functionTemplates/nodejs/';
+const templateDir = 'functionTemplates';
+const runtimeToTemplateDir = {
```
michaeladda: Since we keep `binaris.yml` in the template dir, it might be simpler to just use the runtime name as the directory name and not alter it in run time. It creates a minor duplication of the `python2` and `pypy2` function code, but is simpler and more readable.

lib/functionTemplates/py2/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.py[doc]
```
michaeladda: Since we only currently copy `binaris.yml` and the function file it's ok, but if we'll switch to directory copy, this may hurt the users. Consider keeping `.gitignore` in the `functionTemplates` directory with `*/.binaris` rule for example (same goes for the `node8` template)

lib/binarisYML.js
```diff
@@ -60,13 +49,16 @@ const getFuncName = function getFuncName(binarisConf) {
 };
 
 const checkFuncConf = async function checkFuncConf(funcConf, funcDirPath) {
-  if (!Object.prototype.hasOwnProperty.call(funcConf, fileStr)) {
-    throw new Error(`${binarisConfFile}: function missing required field: <${fileStr}>`);
+  for (const section of [fileStr, entryStr, runtimeStr]) {
+    if (typeof funcConf[section] !== 'string') {
```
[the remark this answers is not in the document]
bergundy: It passes linting..

lib/binarisYML.js
bergundy: I just brought back the code that attempts to read the function and throws the generic fs.readFile error

lib/create.js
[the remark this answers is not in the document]
bergundy: I don't expect this to happen, (like @lavie said, it's just an assertion), do you have a suggestion for a better error message?

lib/create.js
bergundy: FWIU, this is exactly the opposite of what @rylandg says here: https://github.com/binaris/binaris/pull/263/files#diff-da3e06b4a4c89048b62dc5efae8ba65fR34

lib/create.js
bergundy: I don't understand how this will work

lib/functionTemplates/py2/.gitignore
bergundy: I just copied from the node template, is the .gitignore there by mistake?

lib/create.js
rylandg: `const templatePath = path.join(__dirname, templateDir, template);`
`await fse.copy(path.join(__dirname, templateDir, template, file)....`

Fixed
`await fse.copy(path.join(templatePath, file)...`

lib/create.js
rylandg: It's definitely not the opposite. I explicitly said,

> If each runtime had it's own template dir I would suggest hardcoding, but in lieu of that I would say have them all share a root binaris.yml template and update accordingly.

After reading @michaeladda opinion above I believe he put my position into more eloquent words. Each runtime should probably get it's own directory.

lib/create.js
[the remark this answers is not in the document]
rylandg: Not directly but definitely `try reinstalling via npm` or something along those lines might make it better if added. I'm not sure we want users trying to fix this themselves.

lib/functionTemplates/py2/.gitignore
rylandg: It wasn't an issue before because we had a single runtime. I think @michaeladda is suggesting to unify things so `binaris.yml` is handled in a similar fashion to the `.gitignore`.

test/spec.yml
```diff
@@ -201,35 +232,41 @@
...
         err: Function names cannot be longer than 58 characters.
         exit: 1
     -   in: bn create
-        err: "*Not enough non-option arguments: got 0, need at least 1"
+        err: "*Not enough non-option arguments: got 0, need at least 2"
+        exit: 1
+    -   in: bn create node8
+        err: "*Not enough non-option arguments: got 1, need at least 2"
+        exit: 1
+    -   in: bn create abcd
```
michaeladda: can't find a spec for `bn create java8 hello`?
(which will verify the `No template for runtime xxx` error)

lib/create.js
bergundy: Oh sorry @rylandg I misread your comment..

test/spec.yml
bergundy: Added test for cpp4096

PR #276, opened 2018-04-24
Always log raw response (at debug level)

rylandg:
Tested with spec tests. Keeping this simple just to have an answer to annoying debug problems ASAP.

@arielshaqed I'm not opinionated about the exact string so let me know if you have something better in mind.


rylandg:
@arielshaqed Np, changes are made.

PR #279, opened 2018-04-27
Detect if colors are supported, not just TTY

rylandg:
Tested by manually causing errors and verifying colors work. Then using CLI spec and ensuring non-TTY tests pass.


rylandg:
@arielshaqed 
https://github.com/binaris/binaris/issues/283

I checked output by setting a bad endpoint and redirecting to `out.txt`. The colors are indeed not anywhere to be found :)

Issue #183, opened 2017-11-19
invoke --data is unclear


lavie:
The --help does not make it clear how to use --data. Does it expect a string? What format?
Our help should make things obvious. give examples. Show how to supply the param, and show how to consume it inside the function.

rylandg:
https://github.com/binaris/binaris/pull/280 should be enough to at least close this ticket. If we want more specifics we can open a new one.

PR #280, opened 2018-04-27
Add invoke examples

rylandg:
We are lacking any examples which causes issues like https://github.com/binaris/binaris/issues/183.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -93,7 +93,16 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         alias: 'd',
         describe: 'Data to send with invocation',
         type: 'string',
-      });
+      })
+      .example(
+`  // invoke a function
+  bn invoke foo
+
+  // invoke using JSON file data
+  bn invoke foo --json ./path/to/myfile.json
+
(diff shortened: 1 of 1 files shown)


rylandg:
@arielshaqed good point, will use superior style in the future and fix existing usage.

cli.js
```diff
@@ -93,7 +93,16 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
         describe: 'Data to send with invocation',
         type: 'string',
-      });
+      })
+      .example(
+`  // invoke a function
+  bn invoke foo
+
+  // invoke using JSON file data
+  bn invoke foo --json ./path/to/myfile.json
```
lavie: I still find this confusing...

`json` does not convey the fact that it's a file. at first reading, each of these options might signify "from file" or "from string".

Maybe `--json-file`?

rylandg: I'll change it.

lavie: btw, here's my proposal for all the options:

```
--json '[ 1, 2, 3 ]'
--json-file ./test.json
--data 'bare string'
```
WDYT?

PR #285, opened 2018-04-28
Fix login dialog information

rylandg:
Tested with CLI spec tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "5.0.0",
+  "version": "5.0.1",
   "description": "Binaris SDK & CLI",
   "main": "index.js",
   "bin": {
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -69,7 +69,7 @@
             If you don't have a key, head over to https://binaris.com to request one
             *? API Key: *
             *Authentication Succeeded*
-              (use "bn create" to create a template function in your current directory)
+              (use "bn create node8 hello" to create a Node.js template function in your CWD)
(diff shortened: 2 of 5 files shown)


test/spec.yml
```diff
@@ -68,7 +68,7 @@
             If you don't have a key, head over to https://binaris.com to request one
             *? API Key: *
             *Authentication Succeeded*
-              (use "bn create" to create a template function in your current directory)
+              (use "bn create <runtime> <function>" to create a template function in your current directory)
```
michaeladda: IMO this is not clear. either have a working line such as `bn create node8 hello`, or `bn create node8|python2|pypy myFunction`, or instruct to look at the help of the next step, such as `bn create -help`. Using `<runtime>` is not helpful for a user who does not know which runtimes we support. Either way it has to be simple

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "3.0.1",
+  "version": "3.0.3",
```
michaeladda: this is already conflicted

rylandg:
@michaeladda I've updated this. I rebased to get the versioning correct.

rylandg:
@michaeladda 

1. I did omit it, my bad

2. Did you read https://semver.org/? Because the fact our current version is 4.x.x and we still have many potential breaking changes down the road shows that we aren't really adhering well.

rylandg:
@michaeladda this was updated. Any new feedback?

rylandg:
@michaeladda ack

PR #287, opened 2018-04-29
Initial implementation of `bn perf`.

lavie:
Very few options so far, just how many requests and concurrency.
Tested manually, not yet automatically.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -106,7 +106,7 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   }, async (argv) => {
     await handleCommand(argv, invokeHandler);
   })
-  .command('perf <function> [options]', 'Measure invocation latency', (yargs0) => {
+  .command('perf <function> [options]', 'Measure invocation latency (experimental)', (yargs0) => {
     yargs0
       .usage('Usage: $0 perf <function> [options]')
       .positional('function', {
--- a/cli.js
+++ b/cli.js
@@ -117,7 +117,7 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         alias: 'n',
         describe: 'Number of invocations to perform',
         type: 'number',
-        default: 100,
+        default: 5000,
       })
       .option('concurrency', {
         alias: 'c',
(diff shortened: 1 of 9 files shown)


lavie:
I think we should merge this even before automatic testing, since it will be useful for snap testing & demoing.

cli.js
```diff
@@ -106,6 +106,38 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   }, async (argv) => {
     await handleCommand(argv, invokeHandler);
   })
+  .command('perf <function> [options]', 'Measure invocation latency (experimental)', (yargs0) => {
```
arielshaqed: `yargs0`?  Is that a JS standard for this sort of usage?

lib/index.js
```diff
@@ -132,6 +134,43 @@ const invokeHandler = async function invokeHandler(options) {
...
+  logger.info(`Running performance test on function ${options.function}.
+Executing ${options.maxRequests} invocations with ${options.concurrency} "thread${options.concurrency > 1 ? 's' : ''}".
+Stand by for results...
+`);
+  const report = await perf(options.function, options.maxRequests, options.concurrency);
+  logger.info('Perf summary');
+  logger.info('============');
+
+  logger.info(columnify({
+    'Total time': `${report.totalTimeSeconds.toFixed(1)} ms`,
```
arielshaqed: I don't understand the units here: `totalTimeSeconds` is seconds, the suffix `ms` is milliseconds?

lavie: :) fixed.

test/spec.yml
```diff
@@ -12,6 +12,7 @@
                 bn deploy <function> [options]            Deploys a function to the cloud
                 bn remove <function> [options]            Remove a previously deployed function
                 bn invoke <function> [options]            Invoke a Binaris function
+                bn perf <function> [options]              Measure invocation latency (experimental)
```
michaeladda: the spec should also have an actual use of this command

lavie:
@michaeladda please approve so it can be merged. already published to npm.

test/spec.yml
lavie: #293

PR #291, opened 2018-04-29
Revert feature-tags-complete

bergundy:


bergundy:
This reverts #290 which won't work since we rolled back spice

PR #295, opened 2018-05-02
Build job for running spec tests against a realm

lavie:
once approved and merged, I will update the job to use master as the jenkinsfile source: https://ci.binaris.io/job/cli/job/spec-test/

Tested manually by running the job from this branch on private realms and testing.

### DIFF
--- a/spec-tests.Jenkinsfile
+++ b/spec-tests.Jenkinsfile
@@ -1,0 +1,51 @@
+node {
+    checkout scm;
+    properties([
+            parameters([
+                stringParam(
+                    description: 'stable tag of binaris/binaris to use for testing',
+                    name: 'tag'
+                    ),
+                stringParam(
+                    description: 'realm',
+                    name: 'realm'
+                    ),
+            ])
+    ]);
+    if (!params.tag) { error('tag not defined'); }
+    if (!params.realm) { error('realm not defined'); }
+    currentBuild.description = "${params.realm} ${params.tag}"
+
+    ansiColor('xterm') {
+        def realm = params.realm;
+        def tag = params.tag;
+
+        timeout(time: 10, unit: 'MINUTES') {
+            withEnv(['CI=true', "tag=${tag}", "realm=${realm}"]) {
+                withCredentials([string(credentialsId: 'jenkins-binaris-api-key', variable: 'BINARIS_API_KEY')]) {
+                   timestamps {
+                       stage("install api key") {
+                               build(
+                                job: "/create-api-key",
+                                wait: true,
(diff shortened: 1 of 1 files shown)


lavie:
verified that it runs against the realm specified by build param using loggly

spec-tests.Jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+        def realm = params.realm;
+        def tag = params.tag;
+
+        timeout(time: 10, unit: 'MINUTES') {
+            withEnv(['CI=true', "tag=${tag}", "realm=${realm}"]) {
+                withCredentials([string(credentialsId: 'jenkins-binaris-api-key', variable: 'BINARIS_API_KEY')]) {
+                   timestamps {
+                       stage("install api key") {
+                               build(
+                                job: "/create-api-key",
```
rylandg: Glad this job is already of use :)

spec-tests.Jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+                                    string(name: 'BUILD_TAG', value:tag),
+                                    string(name: 'API_KEY', value:"${BINARIS_API_KEY}")
+                                ]
+                             )
+                       }
+                       stage("spec tests") {
+                           withEnv(["tag=${BUILD_TAG}"]) {
+                               // make test tags binaris/binaris with ${tag}, and we don't want to
+                               // somehow override a stable tag with the result of this build.
+                               withCredentials([string(credentialsId: 'realm-config-url', variable: 'REALM_CONFIG_URL')]) {
```
rylandg: We should really modularize the endpoint generation given realm `x` because we do this quite a bit.

lavie:
please some approve this. @rylandg @arielshaqed

spec-tests.Jenkinsfile
```diff
@@ -0,0 +1,59 @@
...
+    ]);
+    if (!params.tag) { error('tag not defined'); }
+    if (!params.realm) { error('realm not defined'); }
+    currentBuild.description = "${params.realm} ${params.tag}"
+
+    ansiColor('xterm') {
+        def realm = params.realm;
+        def tag = params.tag;
+
+        timeout(time: 10, unit: 'MINUTES') {
```
arielshaqed: I'd rather have each of the subjobs have its own timeout.  It doesn't offer the flexibility that allocating a bunch of time to the entire process -- but I would argue that flexibility isn't needed.  OTOH, timeouts (only) on the subjobs mean it's easier to reconfigure a job when we know it's faster/slower.

PR #298, opened 2018-05-03
Remove sleep calls between deploy & invoke

rylandg:
Tests currently fail so I think we should wait until they pass. Otherwise this patch is sound.


rylandg:
@michaeladda nice catch! Fixed

The tests are currently failing with `515` errors after `invoke`, ironically.

PR #307, opened 2018-05-21
bn list

lavie:
@rylandg this is just a PR into the stable branch `feature-list`, so that future PR won't have to re-review every commit in this effort.
please merge based on CR, not build passing (it doesn't, yet)


lib/index.js
```diff
@@ -105,8 +109,8 @@ const deployHandler = async function deployHandler(options) {
  * @param {object} options - Command line options.
  */
 const removeHandler = async function removeHandler(options) {
-  const meta = await gatherMeta(options);
-  await remove(meta.name, meta.path);
+  const meta = await gatherMeta(options, false);
+  await remove(meta.name);
```
rylandg: I thought we decided to avoid `meta` for this call altogether?

lib/list.js
```diff
@@ -0,0 +1,18 @@
+const { getAPIKey } = require('./userConf');
+const { list } = require('../sdk');
+
+/**
+ * Run performance measurements on deployed function
+ *
+ * @param {string} funcName - name of the function to invoke
```
rylandg: This was my bad, I copied the doc from `perf` as a template and didn't update it.

sdk/list.js
```diff
@@ -0,0 +1,16 @@
+const urljoin = require('urljoin');
+const rp = require('request-promise-native');
+
+const { getDeployEndpoint } = require('./config');
+
+const list = async function list(apiKey, endpoint = getDeployEndpoint()) {
+  const listURLBase = `https://${endpoint}`;
+  const listOptions = {
+    url: urljoin(listURLBase, 'v2', 'functions', apiKey),
```
rylandg: I guess your pet peeve is relevant here :), my bad

PR #308, opened 2018-05-21
List spec

lavie:
list tests pass


lib/list.js
```diff
@@ -2,13 +2,20 @@ const { getAPIKey } = require('./userConf');
 const { list } = require('../sdk');
 
 /**
- * Run performance measurements on deployed function
+ * Lust currently deployed functions.
```
rylandg: Those functions must be very enticing :)

PR #309, opened 2018-05-21
bn list feature

lavie:


lib/index.js
```diff
@@ -1,9 +1,10 @@
 const deploy = require('./deploy');
 const create = require('./create');
 const invoke = require('./invoke');
-const perf = require('./perf');
+const list = require('./list');
```
bergundy: much sort!

test/spec.yml
```diff
@@ -20,7 +21,17 @@
...
                 --help, -h  Show help  [boolean]*
+- test: Test list help output
+  steps:
+    -   in: bn list --help
+        out: |-
+              Usage: bn list [options]
 
+              Options:
+                --version   Show version number  [boolean]
+                --path, -p  Use directory dir. "create" will create this directory if needed.  [string]
```
bergundy: Just clarifying, why do we need `--path` here?

PR #312, opened 2018-05-24
allow variance in column width in bn list test

lavie:


lavie:
will add more exact json coverage with jq

lavie:
on second thought, jq not needed. wildcards should suffice
closing this and re-opening with a feature branch

PR #314, opened 2018-05-24
bn list test fixes

lavie:


test/spec.yml
```diff
@@ -137,12 +137,12 @@
...
-             FUNCTION                  LAST DEPLOYED*
-             listtest1                 201*-*-*T*:*:*Z*
-             listtest2                 201*-*-*T*:*:*Z*
+             FUNCTION                 *LAST DEPLOYED*
+             listtest1                *201*-*-*T*:*:*Z*
+             listtest2                *201*-*-*T*:*:*Z*
     -   in: bn list --json
         out: |-
-            *[{"name":"listtest1","lastDeployed":"[20*-*-*T*:*:*Z]"},{"name":"listtest2","lastDeployed":"[20*-*-*T*:*:*Z]"}]*
+            *[*{"name":"listtest1","lastDeployed":"[20*-*-*T*:*:*Z]"}*,*{"name":"listtest2","lastDeployed":"[20*-*-*T*:*:*Z]"}*]*
```
michaeladda: I think either the second or third added asterisk is superfluous (surrounding the comma).

test/spec.yml
```diff
@@ -137,12 +137,12 @@
     -   in: bn deploy listtest2
     -   in: bn list
         out: |-
-             FUNCTION                  LAST DEPLOYED*
-             listtest1                 201*-*-*T*:*:*Z*
-             listtest2                 201*-*-*T*:*:*Z*
+             FUNCTION                 *LAST DEPLOYED*
+             listtest1                *201*-*-*T*:*:*Z*
```
michaeladda: the asterisk is not possitioned correctly imho.
i think the last asterisk in the line should be doubled in all three lines, which will allow extra lines between.

michaeladda: ok, i'm wrong it seems. we changed the asterisk to be multiline a while ago.

PR #315, opened 2018-05-24
Publish cli

rylandg:
The only testing I can really do is the bad case (which I did), make without NPM_TAG.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ test: build
 			bash -c "cd /home/dockeruser/binaris && npm run test"
 
 .PHONY: publish
-publish: build require-npm-creds
+publish: build require-npm-creds require-npm-tag
 		export tag=$(tag)
 		$(DOCKER) run                                                                                              \
 			--rm                                                                                                     \
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,9 @@ test: build
 
 .PHONY: publish
 publish: build require-npm-creds require-npm-tag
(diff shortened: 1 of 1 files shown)


rylandg:
@michaeladda I edited the original PR note to say the testing. It wasn't tested extensively. Any recommendations on viable testing without actually publishing?

Makefile
```diff
@@ -45,19 +45,26 @@ test: build
 			bash -c "cd /home/dockeruser/binaris && npm run test"
 
 .PHONY: publish
-publish: build require-npm-creds
+publish: build require-npm-creds require-npm-tag
+		version=$(shell cat package.json | jq -r ".version")
+		git tag $(version)
```
michaeladda: shouldn't NPM_TAG be here too?
otherwise you assume single npm tag per version and the mapping is not 1 to 1

rylandg: We've never discussed the mapping not being 1:1. We use `NPM_TAG` in the target you've commented on.

rylandg: Oh I see you mean instead of the `version`. I guess that makes sense, this is just the way we do it elsewhere.

rylandg:
So I've made some changes and tested this manually. I still have not tested it with a full end-to-end because `publish` doesn't work the way you described. Basically we will need a second job which does `npm dist-tag` so we can add tags to existing versions. 

I did manually verify the git portion works and the correct NPM tag propagate into the environment.

PR #316, opened 2018-05-30
This will be auto-built by docker hub and published as `binaris/bn`. Only from master branch.

lavie:


lavie:
This is convenience tool, in order to be able to run `bn` CLI where node.js is not installed (e.g. on a build machine that's fully dockerized).

lavie:
After merging, need to update the build settings here to publish master as latest: https://hub.docker.com/r/binaris/bn/~/settings/automated-builds/

Future work (nice to have): have docker images version tags that match npm version.

lavie:
I've also added tagging. So if you push a tag to git like `4.1.10` it till become `binaris/bn:4.1.10` on docker hub.
Master still published as `latest`. If we want another branch to be the one that turns into `latest`

arielshaqed:
Also please do consider using a separate branch for pushing to DockerHub.  Then we can publish `npm` and DockerHub latest together.  Otherwise it will be confusing to get a version from DockerHub -- even the `package.json` version number is not sufficient to identify what exactly is running.

README.md
```diff
@@ -2,7 +2,7 @@
 
 **[Binaris](https://www.binaris.com/)** is a fast, low-latency FaaS (Function as a Service) platform. With our performance and scale, you can run real production workloads on Node.js.
 
-### <a name="up-and-running"></a>Getting up and running in seconds
+## Getting up and running in seconds
```
arielshaqed: Yay!  (We get the anchor anyway, `#getting-up-and-running-in-seconds`.)

PR #319, opened 2018-06-03
No more binaris.yml for 'bn logs'

rylandg:
tested via spec tests and manual invocations


rylandg:
@lavie this will need re-approval because I had to manually rebase this branch on an updated master.

PR #321, opened 2018-06-04
Feature local jenkins

rylandg:
I'm not sure if this is the best way or even the initial way to approach the problem. It's kind of frustrating to have `(local)` as a prefix on all our local tests, but since we don't use Ava in a traditional way this may be one of our only options. I would love to know if this approach seems reasonable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,18 @@ test: build
 			$(cli_envs) $(DOCKER_IMAGE):$(tag)            \
 			bash -c "cd /home/dockeruser/binaris && npm run test"
 
+.PHONY: test_local
+test_local: build
+		export tag=$(tag)
+		$(DOCKER) run                                     \
+			--rm                                          \
+			--privileged                                  \
+			--user root                                   \
+			-v /var/run/docker.sock:/var/run/docker.sock  \
+			$(cli_envs) $(DOCKER_IMAGE):$(tag)            \
+			bash -c "cd /home/dockeruser/binaris && npm test -- -m '(local)*'"
+
+
 .PHONY: publish
 publish: build require-npm-creds
 		export tag=$(tag)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,6 +26,10 @@ node {
                     echo 'Linting'
                     sh "make lint"
                 }
+                stage('Test local') {
+                    echo 'Running local tests'
+                    sh "make test_local"
+                }
                 stage('Tag') {
                     echo 'Tagging local images';
                     sh "make tag"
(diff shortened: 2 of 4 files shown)


rylandg:
@michaeladda I figured that it was more worthwhile to preserve the unity of the `spec`. This is due to us already separating the tests between `spec` and `cases`. Which way do you prefer (I'm indifferent)?

michaeladda:
I'm pretty indifferent to the names too. As indicated above, `unit`, `component`, `integration`, `spec`, etc are all good and meaningful names. You know all the test cases so pick the most appropriate categories.

rylandg:
@michaeladda ok I will revise this patch tomorrow by first creating a PR which splits the tests appropriately.

rylandg:
@michaeladda unfortunately I don't believe we can do this with vanilla `ava`. The issue stems from our very non-traditional testing setup which doesn't feed directories to the `ava` test runner. Instead we generate those tests programmatically by reading in `.yml` files. From `avas` point of view, all of the tests are being executed in a single directory by a single file.

I could split up the tests as we discussed but instead  of directories use names only `unit.yml` etc. Then during runtime I could programmatically prepend every test we create from a specific file (ie: every test in `unit.yml` will have `unit <rest of Test name>`) and run them through `ava`. The downside is that we still can't limit tests by file path/directory structure and will still rely on `match=` semantics. I don't necessarily see this as a huge improvement over my initial set of changes so I'll wait to see what you think.

Alternatively we can jury-rig `ava` and see if we can't borrow the command line input they receive.

rylandg:
@michaeladda waiting for response from you.

Makefile
```diff
@@ -44,6 +44,18 @@ test: build
 			$(cli_envs) $(DOCKER_IMAGE):$(tag)            \
 			bash -c "cd /home/dockeruser/binaris && npm run test"
 
+.PHONY: test_local
+test_local: build
+		export tag=$(tag)
```
michaeladda: DRY with `test`
Maybe add TEST := ...

Makefile
```diff
@@ -44,6 +44,18 @@ test: build
...
+		$(DOCKER) run                                     \
+			--rm                                          \
+			--privileged                                  \
+			--user root                                   \
+			-v /var/run/docker.sock:/var/run/docker.sock  \
+			$(cli_envs) $(DOCKER_IMAGE):$(tag)            \
+			bash -c "cd /home/dockeruser/binaris && npm test -- -m '(local)*'"
```
michaeladda: shouldn't this be npm run test

you can add `-- -m ...` as variable to the test local target and keep it unified as well

michaeladda:
@rylandg This is a very old PR, which was approved long ago and became stale once again. Please either close or resolve and merge.

rylandg:
After some discussion I've decided to close this PR as it's drifted far away from its original intention. The issue is still open and I'll create an updated fix at a later time.

PR #322, opened 2018-06-04
Don't show API_KEY during login

rylandg:
@lavie this is something you brought up so I wanted to tell you why the solution is not optimal. It's not simple to only mask a portion of the text segment (ie: *********KEY) using `inquirer.js`. Allegedly there is a field called `transformer` that should allow a custom function to modify the incremental input but it simply does not work when I define it. I figured this is the best immediate solution.


rylandg:
@lavie  isn't that what this patch does? It just masks the entire entry.

PR #323, opened 2018-06-07
BDSM

lavie:


sdk/deploy.js
```diff
@@ -62,6 +62,34 @@ const deployCode = async function deployCode(deployURLBase, apiKey, tarPath) {
  * @return {object} - response of tag operation
  */
 const deployConf = async function deployConf(deployURLBase, apiKey, funcName, funcConf) {
+  const { env } = funcConf;
+  if (env) {
+    for (const key of Object.keys(env)) {
+      if (env[key] === '') {
+        logger.error(`
```
arielshaqed: What if I need to set an empty environment variable?  E.g. shell `IFS` has a different default behaviour when unset.

lavie: There's hardly a reason to pass empty strings to a FaaS function. In any case, it's not possible ATM. If users ask for it we can add it (requires escaping this special case in Dynamo, which refuses to accept empty string values)

arielshaqed: 1. People will do it; it's a special case otherwise.
2. If it's a DynamoDB limitation, `spice` should enforce it.  Otherwsie it becomes harded to fix.
3. I do understand the DynamoDB limitation / special case / bug.  Why are we storing the `env` flatly inside DynamoDB?  E.g. if we stored `JSON.encode(env)`, the empty-string limitation would vanish... plus we'd get nice rich data for user values.

lavie: bottom line, we can deliver this feature today or we can stop and add support for empty strings, which nobody asked for, and not have it done for Optibus Sunday.

I think we're bike shedding here. This is good enough.

michaeladda: 1. I'm not sure I understand why we don't pack the structure, and I think we should, but it can certainly be done later. an issue is a good compromise.
2. This is a public repo, please don't forget it.

lavie: comment redacted. sorry.

PR #324, opened 2018-06-07
Update README.md

lavie:


README.md
```diff
@@ -84,6 +84,64 @@ bn invoke hello
...
+functions:
+  hello:
+    file: function.js
+    entrypoint: handler
+    runtime: node8
+    env:
+      SOME_API_KEY: XXXXXXXXXXX
+```
+
+If a parameter is sensitive and you don't want to accidentally commit it to SCM, you can leave out the value in `binaris.yml`:
```
arielshaqed: Remove `accidentally`?

README.md
```diff
@@ -84,6 +84,64 @@ bn invoke hello
...
+functions:
+  hello:
+    file: function.js
+    entrypoint: handler
+    runtime: node8
+    env:
+      SOME_API_KEY:
+```
+
+And pass it as an environment variable to the `deploy` command:
```
arielshaqed: Example from 2 paras down should go here I think.

PR #331, opened 2018-08-14
Feature better errors

rylandg:
Trying to improve the general user experience. This is just an initial PR for feedback.

### DIFF
--- a/lib/handleError.js
+++ b/lib/handleError.js
@@ -15,8 +15,6 @@ module.exports = {
       throw new Error(translateErrorCode(response.body.errorCode));
     } else if (get(response, 'error')) {
       throw response.error;
-    } else {
-      throw new Error(translateErrorCode('ERR_NO_BACKEND'));
     }
   },
 };
--- a/lib/handleError.js
+++ b/lib/handleError.js
@@ -13,6 +13,8 @@ module.exports = {
     }
     if (get(response, 'body.errorCode')) {
       throw new Error(translateErrorCode(response.body.errorCode));
+    } else if (get(response, 'error')) {
+      throw response.error;
     } else {
       throw new Error(translateErrorCode('ERR_NO_BACKEND'));
     }
--- a/sdk/remove.js
+++ b/sdk/remove.js
@@ -21,9 +21,7 @@ const remove = async function remove(funcName, apiKey) {
     const response = await tryRequest(rp.delete(options));
     return { status: response.statusCode, body: response.body };
   } catch (err) {
-    // console.log(JSON.stringify(err));
     if (err instanceof HTTPError) {
-      console.log('http error');
       return { status: err.response.statusCode, body: err.response.body };
     }
     // NOTE: This 'err' returned in the error field is in NodeJS error format
--- a/sdk/remove.js
+++ b/sdk/remove.js
@@ -4,6 +4,7 @@ const get = require('lodash.get');
 
 const { translateErrorCode } = require('binaris-pickle');
 const { getDeployEndpoint } = require('./config');
+const { HTTPError, tryRequest } = require('./httpError');
 
 /**
  * Removes the function from the Binaris cloud.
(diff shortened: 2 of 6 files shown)


sdk/remove.js
```diff
@@ -18,17 +17,17 @@ const remove = async function remove(funcName, apiKey) {
...
-    response = await rp.delete(options);
+    const response = await tryRequest(rp.delete(options));
+    return { status: response.statusCode, body: response.body };
   } catch (err) {
-    throw new Error(translateErrorCode('ERR_NO_BACKEND'));
-  }
-  if (response.statusCode < 200 || response.statusCode >= 300) {
-    if (get(response, 'body.errorCode')) {
-      throw new Error(translateErrorCode(response.body.errorCode));
+    // console.log(JSON.stringify(err));
```
michaeladda: ?

test/cases.yml
```diff
@@ -25,10 +25,10 @@
   steps:
     -   in: bn create node8 naughtyvegetable
     -   in: bn deploy naughtyvegetable
-        err: Could not connect to Binaris backend
+        err: "getaddrinfo ENOTFOUND fake.binaris.neverworks fake.binaris.neverworks:443"
         exit: 1
     -   in: bn remove naughtyvegetable
-        err: Could not connect to Binaris backend
+        err: "Error: getaddrinfo ENOTFOUND fake.binaris.neverworks fake.binaris.neverworks:443"
```
michaeladda: why are the error messages different?
(I don't mean technically why, I mean specification wise. They should be similar)

sdk/remove.js
```diff
@@ -18,17 +17,17 @@ const remove = async function remove(funcName, apiKey) {
...
+    if (err instanceof HTTPError) {
+      console.log('http error');
```
michaeladda: why?!?

and more importantly why create a special case in the first place?

test/cases.yml
rylandg: If you're talking about between `deploy` and `remove` it's because `deploy` uses `request` whereas `remove` uses `request-promise-native`. I could do some specific coercion  on the `deploy` to make it fit with the traditional error but wasn't sure how important it is.

sdk/remove.js
rylandg: This is the way it was implemented by the author of our `sdk/deploy`and I was just trying to maintain uniformity.

lib/handleError.js
```diff
@@ -13,8 +13,8 @@ module.exports = {
     }
     if (get(response, 'body.errorCode')) {
       throw new Error(translateErrorCode(response.body.errorCode));
-    } else {
-      throw new Error(translateErrorCode('ERR_NO_BACKEND'));
+    } else if (get(response, 'error')) {
```
vogre: I see that it's only possible to be used as 
```res.status(404).send('UPGRADE REQUIRED')``` to fill `response.error`, however the contents may be `<!DOCTYPE>...` - a response from a misconfigured server which will dump the HTML into the console.

Comment (rylandg) [reply lib/handleError.js]: @vogre  I don't believe this is the case. We construct the error object in `deploy` and `remove` and explicitly set the `error` field. In the case of us sending something misconfigured I believe that `request` will surface that via normal error handling.

```
  try {
    const deployURLBase = `https://${endpoint}`;
    const codeDigest = await deployCode(deployURLBase, apiKey, tarPath);
    const confWithDigest = Object.assign({}, funcConf, { codeDigest });
    const confDeployResp = await deployConf(deployURLBase, apiKey,
      funcName, confWithDigest);
    return { status: confDeployResp.statusCode, body: confDeployResp.body };
  } catch (err) {
    if (err instanceof HTTPError) {
      return { status: err.response.statusCode, body: err.response.body };
    }
    // NOTE: This 'err' returned in the error field is in NodeJS error format
    return { error: err };
  }
```

PR #333, opened 2018-08-16
Greatly improve error feedback

rylandg:
Tested manually and with spec but won't merge until I've used the branch in active development for a bit.


sdk/deploy.js
```diff
@@ -41,7 +37,7 @@ const deployCode = async function deployCode(deployURLBase, apiKey, tarPath) {
         if (uploadErr) {
           reject(uploadErr);
         } else if (uploadResponse.statusCode !== 200) {
-          reject(new HTTPError(uploadResponse));
+          reject(new APIError(translateErrorCode(uploadResponse.body.errorCode)));
```
vogre: Can we call into `handleError.js` logic here to have similar `errorText` handling?

sdk/handleError.js
```diff
@@ -0,0 +1,40 @@
...
+
+const logger = require('../lib/logger');
+
+const { translateErrorCode } = require('binaris-pickle');
+
+class APIError extends Error {}
+
+async function callAPI(options, type = 'post') {
+  const response = await rp[type]({
+    json: true,
```
michaeladda: json true appears both here and in callers to `callAPI`. This needs some cleanup.
The only options I see being used are url, body, and method. Method already has its special argument (type). Wouldn't passing all of these explicitly be more readable?
Currently, if the caller sets `simple: true` or `resolveWithFullResponse: true` or even `method: 'GET'` it can get very confusing

sdk/handleError.js
```diff
@@ -0,0 +1,40 @@
...
+const { inspect } = require('util');
+const rp = require('request-promise-native');
+
+const logger = require('../lib/logger');
+
+const { translateErrorCode } = require('binaris-pickle');
+
+class APIError extends Error {}
+
+async function callAPI(options, type = 'post') {
```
michaeladda: `callAPI` is a really bad name. It's both generic and does not explain what it does.
How about something like loggedRequest of verboseRP or requestWithNiceErrors, or ...
(you get it)

sdk/deploy.js
rylandg: Sure I'll break down the handleError logic and call the relevant part.

lib/remove.js
```diff
@@ -8,7 +8,8 @@ const { getAPIKey } = require('./userConf');
  */
 const removeCLI = async function removeCLI(funcName) {
   const apiKey = await getAPIKey();
-  return remove(funcName, apiKey);
+  const removeResponse = await remove(funcName, apiKey);
```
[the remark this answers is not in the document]
rylandg: Well in a general sense they aren't the same. The await placement changes the error catching/handling responsibility. In this case I think the difference is unimportant so I'll change it.

sdk/handleError.js
rylandg: Consider it done.

sdk/handleError.js
rylandg: The design was based on this conversation in the CLI channel in slack https://binarisio.slack.com/archives/C724CG1J5/p1534441584000100.

I agree we should definitely make it clear what is being done automatically and what needs to be passed explicitly. That being said if a user wanted JSON to be false couldn't he simply override that himself?

sdk/test/deployTest.js
```diff
@@ -72,19 +70,16 @@ test.serial('Test deploy with bad key (bad-path)', async (t) => {
...
     .reply(403, { errorCode: 'ERR_BAD_KEY' });
 
-  const response = await deploy(testFuncName, someBadKey,
-    testFuncConf, t.context.fakeTarFileName, deployEndpoint);
-
-  t.is(403, response.status);
-  t.is('ERR_BAD_KEY', response.body.errorCode);
+  await t.throws(deploy(testFuncName, someBadKey,
+    testFuncConf, t.context.fakeTarFileName, deployEndpoint),
+    'Invalid API key');
```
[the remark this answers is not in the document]
rylandg: Good point.

test/cases.yml
```diff
@@ -25,10 +25,10 @@
   steps:
     -   in: bn create node8 naughtyvegetable
     -   in: bn deploy naughtyvegetable
-        err: Could not connect to Binaris backend
+        err: "getaddrinfo ENOTFOUND fake.binaris.neverworks fake.binaris.neverworks:443"
         exit: 1
     -   in: bn remove naughtyvegetable
-        err: Could not connect to Binaris backend
+        err: "Error: getaddrinfo ENOTFOUND fake.binaris.neverworks fake.binaris.neverworks:443"
```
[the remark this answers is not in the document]
rylandg: The changes Vlad suggested (and I agreed with) were considerable. The PR you're discussing solved the error surfacing changes in a much different way. 

That being said I'll make sure the errors match.

sdk/test/deployTest.js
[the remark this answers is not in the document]
rylandg: https://github.com/binaris/binaris/issues/334

rylandg:
@michaeladda @vogre I believe all issues have been addressed.

PR #337, opened 2018-09-03
Feature abacus

lavie:


sdk/stats.js
```diff
@@ -0,0 +1,29 @@
...
+  const statsOptions = {
+    url: urljoin(statsURLBase, 'v2', 'metrics', apiKey),
+    json: true,
+    body: {
+      since,
+      until,
+    },
+  };
+  logger.debug('Fetching account usage stats', { statsOptions });
+  const metrics = await rp.get(statsOptions);
```
vogre: Please use 

    const { loggedRequest } = require('./handleError');
    await loggedRequest(opt, 'get');

This will ensure errors (like upgrade request for cli version) are translated correctly to the user.

test/test.js
```diff
@@ -92,9 +92,15 @@ function createTest(rawSubTest) {
     t.context.cleanup = rawSubTest.cleanup;
 
     const createRegTest = function createRegTest(expected) {
-      const escaped = expected.split('*').map(item => regEsc(item));
-      const finalString = `^${escaped.join('[\\s\\S]*')}$`;
-      return new RegExp(finalString);
+      const DIGIT_ESCAPE = 'ESCAPESEQUENCEDIGIT';
+      const SPACE_ESCAPE = 'ESCAPESEQUENCESPACE';
```
vogre: `SPACE_ESCAPE` more appropriate as `ANYCHAR_ESCAPE`?

lib/index.js
```diff
@@ -216,6 +218,30 @@ const logsHandler = async function logsHandler(options) {
...
 
+/**
+ * Retrieve account usage stats
+ *
+ * @param {object} options - Command line options.
+ */
+const statsHandler = async function statsHandler(options) {
+  const metrics = await stats(options.since, options.until);
+  // eslint-disable-next-line no-param-reassign
+  metrics.metrics.forEach((m) => { delete m.account; });
```
vogre: If we bring `lodash`, we can use `omit` to avoid `delete`.

metrics.metrics = metrics.metrics.map((m) => omit(m, 'account'));

lib/index.js
```diff
@@ -216,6 +218,30 @@ const logsHandler = async function logsHandler(options) {
...
+  const columns = ['function', 'metric', 'value', 'since', 'until'];
+  metrics.metrics = sortBy(metrics.metrics, columns);
+  if (options.json) {
+    logger.info(JSON.stringify(metrics));
+  } else {
+    if (metrics.metrics.length === 0) {
+      logger.info('No matching usage stats found');
```
vogre: To make linter happy can be written as one `logger.info`

        logger.info(!metrics.metrics.length ? 'No matching usage stats found' : columnify(metrics.metrics, {
          showHeaders: true,
          columns,
        }));

sdk/stats.js
```diff
@@ -0,0 +1,29 @@
...
+const { getDeployEndpoint } = require('./config');
+
+/**
+ * Retrieves account usage statistics
+ *
+ * @param {string} apiKey - Binaris API key
+ * @param {Date} since - All stats since the date-time
+ * @param {Date} until - All stats until the date-time
+ */
+const stats = async function stats(apiKey, since, until) { // eslint-disable-line consistent-return,max-len
```
vogre: Micro-style comment: the `// eslint-disable-line` seems obsolete.

Also could be written as `async function stats` without `const stats =` to avoid `Foo foo` repetition.

lib/index.js
vogre: As a convention, `delete` implies some mutable state, e.g. using an object as a hash, and here we are discarding the metrics as soon as they are printed.

In this case it's simple to understand the intention from the code, it may only be confusing in the future.

PR #338, opened 2018-09-04
Feature enforce versioning

rylandg:
Tested manually. As you can see causes no real API changes.

### DIFF
--- a/lib/deploy.js
+++ b/lib/deploy.js
@@ -49,11 +49,8 @@ const deployCLI = async function deployCLI(funcName, funcPath, funcConf) {
     },
   });
   const apiKey = await getAPIKey();
-  const { advice } = await deploy(funcName, apiKey, funcConf, funcTarPath);
-  return {
-    advice,
-    body: urljoin(`https://${getInvokeEndpoint()}`, 'v1', 'run', apiKey, funcName),
-  };
+  await deploy(funcName, apiKey, funcConf, funcTarPath);
+  return urljoin(`https://${getInvokeEndpoint()}`, 'v1', 'run', apiKey, funcName);
 };
 
 module.exports = deployCLI;
--- a/lib/deploy.js
+++ b/lib/deploy.js
@@ -49,8 +49,11 @@ const deployCLI = async function deployCLI(funcName, funcPath, funcConf) {
     },
   });
   const apiKey = await getAPIKey();
-  await deploy(funcName, apiKey, funcConf, funcTarPath);
-  return urljoin(`https://${getInvokeEndpoint()}`, 'v1', 'run', apiKey, funcName);
+  const { advice } = await deploy(funcName, apiKey, funcConf, funcTarPath);
+  return {
+    advice,
+    body: urljoin(`https://${getInvokeEndpoint()}`, 'v1', 'run', apiKey, funcName),
+  };
 };
 
 module.exports = deployCLI;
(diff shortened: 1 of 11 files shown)


sdk/deploy.js
```diff
@@ -22,6 +22,7 @@ const deployCode = async function deployCode(deployURLBase, apiKey, tarPath) {
     headers: {
       'Content-Type': 'application/gzip',
       'X-Binaris-Api-Key': apiKey,
+      'X-Binaris-CLI-Version': version,
```
michaeladda: isn't this the sdk version?

lib/deploy.js
```diff
@@ -49,8 +49,11 @@ const deployCLI = async function deployCLI(funcName, funcPath, funcConf) {
     },
   });
   const apiKey = await getAPIKey();
-  await deploy(funcName, apiKey, funcConf, funcTarPath);
-  return urljoin(`https://${getInvokeEndpoint()}`, 'v1', 'run', apiKey, funcName);
+  const { advice } = await deploy(funcName, apiKey, funcConf, funcTarPath);
```
michaeladda: why is this only part of deploy? what about remove, log, list?
If this is deploy only feature, no need to print from common `exceptionWrapper`.
If it isn't then why aren't other commands do so too?

rylandg: The other commands have been updated since I submitted this patch. I'm trying to keep this granular and agile. Once I make the same updates to `picard` etc as I did to `spice` I'll make the same changes to other SDK/CLI commands.

sdk/deploy.js
rylandg: I'll change it to `X-Binaris-Client-Version`.

bergundy: I'd add the header in `loggedRequest`, simply check if `headers` are passed in and add the version header to the ones passed in. (see comment below)

sdk/handleError.js
```diff
@@ -23,14 +25,20 @@ function validateResponse(response) {
...
+    advice: get(response, 'headers.advice'),
+  };
 }
 
 async function loggedRequest(options, type = 'post') {
   const response = await rp[type]({
     json: true,
     simple: false,
     resolveWithFullResponse: true,
+    headers: {
```
bergundy:
```js
const { headers } = options;
```
...
```js
headers: { ...(headers || {}), 'X-binaris-CLI-Version': version },
```

bergundy:
Closes #348

rylandg:
I've removed the advice portion (for now) and changed the header to be `X-Binaris-Client-Version`

sdk/handleError.js
```diff
@@ -32,6 +34,10 @@ async function loggedRequest(options, type = 'post') {
     simple: false,
     resolveWithFullResponse: true,
     ...options,
+    headers: {
+      ...(options.headers || {}),
+      'X-binaris-Client-Version': version,
```
vogre: Headers are case insensitive, but `X-binaris-Client` looks asymmetrical.

rylandg: My bad, nice catch.

rylandg:
@vogre please look again.

vogre:
LGTM, @michaeladda WDYT?

rylandg:
@michaeladda I already made the change Roey requested.

michaeladda:
a little too late now but package version should have been bumped, and a new cli should be published after this.

Issue #336, opened 2018-08-26
commands fail silently because of bad api key


lavie:
If a non-existing API key is used for commands, often the result is just a seemingly healthy response that looks empty.

e.g.
`bn logs`, `bn list` - if the API key does not exist the result is just an empty printout. No word to the user about the api key problem.

rylandg:
First portion fixed here https://github.com/binaris/binaris/pull/339

rylandg:
Surprisingly the rest of this issue `bn logs` is actually due to a backend issue with authentication.

vogre:
`bn logs` returns errors now.

PR #339, opened 2018-09-06
Use `loggedRequest` for list

rylandg:
This should fix the `list` portion of the issue described here https://github.com/binaris/binaris/issues/336. Unfortunately our setup for logging is more complex so that change will arrive separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "5.3.0",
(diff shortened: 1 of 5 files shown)


rylandg:
Change done. I agree about the other areas getting "Error" prefix although this is handled inherently by `handleError.js`. I still need to work with `bn logs` because it uses a much more complicated request format.

PR #340, opened 2018-09-12
Feature log standard

rylandg:
Switch `bn logs` to use the new standardized error handling. Tested with spec and manual attempts

I also considered changing `handleError/validateResponse` so it doesn't return at all. If the reviewer prefers that style I'll make the change.

### DIFF
--- a/sdk/handleError.js
+++ b/sdk/handleError.js
@@ -23,7 +23,7 @@ function validateResponse(response) {
     throw new APIError(`Unexpected response ${response.statusCode}`);
   }
 
-  return response.body || response;
+  return response;
 }
 
 async function loggedRequest(options, type = 'post') {
--- a/sdk/handleError.js
+++ b/sdk/handleError.js
@@ -26,7 +26,7 @@ function validateResponse(response) {
   return response;
 }
 
-async function loggedRequest(options, type = 'post') {
+async function loggedRequest(options, type = 'post', validate = true) {
   const response = await rp[type]({
     json: true,
     simple: false,
@@ -34,7 +34,10 @@ async function loggedRequest(options, type = 'post') {
     ...options,
   });
   logger.debug('raw response', inspect(response, { depth: null }));
-  return validateResponse(response);
+  if (validate) {
+    return validateResponse(response);
+  }
+  return response;
 }
 
 module.exports = {
--- a/sdk/handleError.js
+++ b/sdk/handleError.js
@@ -26,7 +26,7 @@ function validateResponse(response) {
   return response;
 }
 
-async function loggedRequest(options, type = 'post', validate = true) {
+async function loggedRequest(options, type = 'post') {
   const response = await rp[type]({
     json: true,
     simple: false,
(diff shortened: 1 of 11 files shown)


sdk/logs.js
```diff
@@ -29,23 +27,17 @@ const logs = async function logs(functionName, apiKey, follow, startAfter, token
   };
 
   for (let attempt = 1, backoff = 3; attempt <= 3; attempt += 1, backoff *= 2) {
+    // eslint-disable-next-line no-await-in-loop
+    const response = await loggedRequest(options, 'get', false);
```
michaeladda: 1. Why not use the validated version? 
2. Why fail silently after 3 attempts and not alert the user?
(not sure about the multiple attempts in the first place, nor about the backoff ms sleep times (3,6,12) but this is another issue)

sdk/handleError.js
```diff
@@ -23,18 +23,21 @@ function validateResponse(response) {
     throw new APIError(`Unexpected response ${response.statusCode}`);
   }
 
-  return response.body || response;
+  return response;
 }
 
-async function loggedRequest(options, type = 'post') {
```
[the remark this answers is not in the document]
rylandg: Because it was an oversight (by me) to believe that validating a request only means you need `response.body`. In my mind it makes much more sense to just return the whole object and let the caller extract as needed.

sdk/logs.js
rylandg: The validated version would remove the logic you describe in point 2.

I wasn't the author of the logging logic so can't provide insight into the specific choices made. Again most changes I make are with agility in mind, the feature is to standardize the error handling/requests between SDK calls, not improve logging logic.

michaeladda:
@rylandg any update?

rylandg:
@michaeladda I think I was waiting for responses to my answers from you.

sdk/stats.js
```diff
@@ -22,8 +22,7 @@ const stats = async function stats(apiKey, since, until) { // eslint-disable-lin
   };
 
   logger.debug('Fetching account usage stats', { statsOptions });
-  const metrics = await loggedRequest(statsOptions, 'get');
-  return metrics;
+  return getValidatedBody(statsOptions, 'get');
```
michaeladda: why drop the `await`?
it doesn't matter right now, but it's a matter of style and design.
if `getValidatedBody` errors should be handled by `stats` (or maybe augmented) it should have an `await`, otherwise, if it's the caller responsibility, a simple `return` is in place

sdk/logs.js
```diff
@@ -31,21 +29,14 @@ const logs = async function logs(functionName, apiKey, follow, startAfter, token
   for (let attempt = 1, backoff = 3; attempt <= 3; attempt += 1, backoff *= 2) {
     try {
       // eslint-disable-next-line no-await-in-loop
-      const { statusCode, headers, body } = await rp.get(options);
-      logger.debug('logs response', { statusCode, headers, body });
+      const { body, headers } = validateResponse(await loggedRequest(options, 'get'));
```
michaeladda: not as succinct, but I find
```
const response = await loggedRequest(options, 'get');
const { body, headers } = validateResponse(response);
```
more readable

rylandg: Yea I agree, it's more readable. I'll clean it up in a later patch.

sdk/stats.js
rylandg: Well it's not in a try/catch so the await didn't matter anyway. I'm also assuming that's why linting failed (because the await doesn't matter here).

PR #343, opened 2018-09-27
add python3 to supported runtimes

michaeladda:
tested with CI

### DIFF
--- a/lib/create.js
+++ b/lib/create.js
@@ -10,6 +10,7 @@ const runtimeToTemplateDir = {
   python2: 'py2',
   pypy2: 'py2',
   java8: 'java8',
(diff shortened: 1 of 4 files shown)


lib/create.js
```diff
@@ -10,6 +10,7 @@ const runtimeToTemplateDir = {
   python2: 'py2',
   pypy2: 'py2',
   java8: 'java8',
+  python3: 'py2',
```
rylandg: Should be py3

rylandg: Ahhh understood.

rylandg: I thought we did a print from python like we do node

PR #345, opened 2018-10-16
Feature usable errors

rylandg:
Essentially fixes all error output when using the system. This patch was only made possible by the recent libra change which removed our last HTML error response. Now all user-function errors should look nice in addition to errors from the system itself.


sdk/invoke.js
```diff
@@ -21,10 +21,19 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
...
-  const { statusCode, headers, body } = res;
-  logger.debug('Invoke response', { statusCode, headers, body });
-  return res;
+  try {
+    const res = await rp.post(options);
+    const { statusCode, headers, body } = res;
+    logger.debug('Invoke response', { statusCode, headers, body });
+    return res;
+  } catch (err) {
+    const parsedError = JSON.parse(err.error);
```
michaeladda: this can throw too, and does not account all possible errors.
the user function can return errors on its own which do not conforms to this pattern.
the code must validates the schema completely before assuming it is a binaris internal error.
(and print user originated errors as well, of course)

sdk/invoke.js
```diff
@@ -21,10 +21,19 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
...
+    if (parsedError.stack) {
+      throw new Error(`${parsedError.error}\n${parsedError.stack}`);
```
michaeladda: request_id?

sdk/invoke.js
```diff
@@ -21,10 +21,25 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
...
+  } catch (err) {
+    let parsedError;
+    try {
+      parsedError = JSON.parse(err.error);
+    } catch (nestedErr) {
+      throw new Error(err.error || err);
+    }
+    if (parsedError.stack && (parsedError.error || parsedError.message)) {
+      throw new Error(`${parsedError.error || parsedError.message}\n${parsedError.stack}`);
+    } else if (parsedError.message && parsedError.request_id) {
```
michaeladda: why `else` here and not in line 41. please be consistent.
(our common style is to avoid the else)

sdk/invoke.js
```diff
@@ -21,10 +21,25 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
...
+    logger.debug('Invoke response', { statusCode, headers, body });
+    return res;
+  } catch (err) {
+    let parsedError;
+    try {
+      parsedError = JSON.parse(err.error);
+    } catch (nestedErr) {
+      throw new Error(err.error || err);
+    }
+    if (parsedError.stack && (parsedError.error || parsedError.message)) {
```
michaeladda: consider naming `(parsedError.error || parsedError.message)` as it's used twice

test/spec.yml
```diff
@@ -417,6 +418,140 @@
...
+    -   in: |-
+            sed -i.bak '6s/.*/        throw new java.lang.Error("this is very bad");/' src/main/java/ExampleFunction.java
+    -   in: gradle jar
+    -   in: bn deploy doomedtofailjava8
+        out: |-
+            Deployed function to *
+              (use "bn invoke doomedtofailjava8" to invoke the function)
+    -   in: bn invoke doomedtofailjava8
+        err: |-
+            <html>
```
michaeladda: This is ugly. Please create an issue for it.
I don't want to delay this PR, but it must be handled.

sdk/invoke.js
```diff
@@ -21,10 +21,25 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
...
+    const res = await rp.post(options);
+    const { statusCode, headers, body } = res;
+    logger.debug('Invoke response', { statusCode, headers, body });
+    return res;
+  } catch (err) {
+    let parsedError;
+    try {
+      parsedError = JSON.parse(err.error);
+    } catch (nestedErr) {
+      throw new Error(err.error || err);
```
michaeladda: Should probably be combined with the default throw after the ifs, and should consider adding text here (or in the internal system errors) to differentiate the cases from user perspective.
Is this covered by the spec test below?

sdk/invoke.js
rylandg: Will do.

sdk/invoke.js
rylandg: Definitely could be merged. It is tested by the cases below such as with Java8.

sdk/invoke.js
rylandg: Can you explain the rationale behind not using else? I can maybe understand if this were C and we had to think of branch prediction or something.

test/spec.yml
rylandg: I assume you mean our bolt response is ugly correct?

PR #346, opened 2018-10-17
Feature more spec errors

rylandg:
Tested with the spec...

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -702,6 +702,11 @@
     -   in: bn list
         err: "Error: Invalid API key"
         exit: 1
+    -   in: BINARIS_API_KEY=boguskey bn invoke perplexingpersimmon
+        err: |-
+            Unauthorized request. Make sure you're passing the right API key.
+            request_id: *
+        exit: 1
 
 - test: Deploy of 200MB function succeeds
   setup:
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -427,6 +427,30 @@
             request_id: *
         exit: 1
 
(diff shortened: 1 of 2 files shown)


test/spec.yml
```diff
@@ -427,6 +427,30 @@
...
+           exports.handler = async (body, req) => {
+             const msleep = ms => new Promise(resolve => setTimeout(resolve, ms));
+             await msleep(100000);
+           };
+           EOL
+    -   in: bn deploy timeoutforsure
+    -   in: bn invoke timeoutforsure
+        err: |-
+            <html>
+            <head><title>504 Gateway Time-out</title></head>
```
michaeladda: please open issues for all these unfriendly errors you encounter (these are generated by libra for example)

test/test.js
```diff
@@ -70,6 +70,10 @@ function limitedTest(name, testFn) {
   return test(name, async t => limiter(async () => testFn(t)));
 }
 
+function stripText(origText) {
+  return strip(origText.split('\r').join('').slice(0, -1));
```
michaeladda: why split and splice and not a simple trimming `replace(/\r+/g, '')`?

test/spec.yml
rylandg: I know it's a Libra error...

test/test.js
rylandg: Why use regex when not needed?

michaeladda: because replace is more readable than split & join?
(it's also faster, but this is not a real consideration here)

PR #347, opened 2018-10-17
Feature error cleanup

rylandg:
More cleanup from the errors patch. Tested with spec.

### DIFF
--- a/sdk/invoke.js
+++ b/sdk/invoke.js
@@ -30,13 +30,13 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
     let parsedError;
     try {
       parsedError = JSON.parse(err.error);
+      if (parsedError.stack && (parsedError.error || parsedError.message)) {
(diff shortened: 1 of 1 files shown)


sdk/invoke.js
```diff
@@ -30,13 +30,15 @@ const invoke = async function invoke(funcName, apiKey, funcData) {
     let parsedError;
     try {
       parsedError = JSON.parse(err.error);
+      const errorOrMsg = parsedError.error || parsedError.message;
+      if (parsedError.stack && errorOrMsg) {
+        throw new Error(`${errorOrMsg}\n${parsedError.stack}`);
```
michaeladda: this throw will be caught in line 40, rendering the error useless.
You probably want to rewrite this block with try parse and if parsedError instead to get to a common error at the borrom

rylandg:
Believe it's fixed.

PR #349, opened 2018-10-21
Increase concurrency.

arielshaqed:
System should be (much) better protected against concurrent deploys.
Start exercising it.  Note that during full-system tests 2 other repos
_also_ deploy functions, concurrently, and this is unprotected by
throttling in this repo.


arielshaqed:
> Ideally, we should remove this altogether

Yes.  The system should protect itself from overloads -- even "just" by clamping requests.

PR #350, opened 2018-10-21
test that deploys and removals get logged to picard

lavie:
must be merged together with 
https://github.com/binaris/spice/pull/458

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -312,7 +312,7 @@
         out: |-
           *[201*-*-*T*:*:*Z] Function gulliblezebra deployed (version digest:*)*
           *[201*-*-*T*:*:*Z] Hello World!*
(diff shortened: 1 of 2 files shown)


test/spec.yml
```diff
@@ -297,6 +297,7 @@
         out: |-
             Created function * in /home/dockeruser/test
               (use "bn deploy gulliblezebra" to deploy the function)
+    -   in: sed -i '2 a \  console.log(`Hello ${name}!`);' function.js
```
rylandg: I would replace this with 
```
    -   in: |-
           cat > function.js << EOL
           exports.handler = async (body, req) => { console.log(`Hello ${name}!`) };
           EOL
```

Far more readable.

lavie: done. please have a second look.

PR #354, opened 2018-10-23
Support env var interpolation in yaml

rylandg:
Tested manually and also validated via the spec.

### DIFF
--- a/sdk/deploy.js
+++ b/sdk/deploy.js
@@ -55,6 +55,7 @@ const deployCode = async function deployCode(deployURLBase, apiKey, tarPath) {
  * @return {object} - response of tag operation
  */
 const deployConf = async function deployConf(deployURLBase, apiKey, funcName, funcConf) {
+  const interpolRegex = new RegExp(/^\$[a-zA-Z_]{1,}[a-zA-Z0-9_]{0,}/);
   const { env } = funcConf;
   if (env) {
     for (const key of Object.keys(env)) {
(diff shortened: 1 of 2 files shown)


lavie:
Just a note that it would be far more valuable to be able to do interpolation all over the YAML, not just in the value parts of the env field. e.g. function name should be able to have envars in it, which would automagically give users the ability to do different "Stages" and stuff, even if we don't implement that ourselves. The flexibility is incredible.

sdk/deploy.js
```diff
@@ -55,6 +55,7 @@ const deployCode = async function deployCode(deployURLBase, apiKey, tarPath) {
  * @return {object} - response of tag operation
  */
 const deployConf = async function deployConf(deployURLBase, apiKey, funcName, funcConf) {
+  const interpolRegex = new RegExp(/^\$[a-zA-Z_]{1,}[a-zA-Z0-9_]{0,}/);
```
lavie: We should be more clear about the supported syntax for interpolation.
`$THIS` - sure
`${WHAT_ABOUT_THIS}`?
What would a user expect? How is this documented (should probably be part of this PR).

lavie: I worry about this home-grown implementation of envar expansion. What about the various edge cases? How do you escape `$`? I also think `regex` is a "now you have two problems" type of solution to this problem.

Maybe try to find a NPM package that handles this? e.g. this impl seems to be a bit more robust: https://github.com/danday74/envsub/blob/master/js/envsub-parser.js
If you can't find a good one and you decide to implement it yourself, you should probably publish it to NPM publicly. Useful.

Alternatively, use a proper template engine, like handlebars. While the syntax won't be the same as the familiar SH syntax for interpolation, it's still a widely known syntax. And of course the benefit is that we don't roll our own parser.

rylandg:
Closing this until I've found the appropriate libraries.

PR #355, opened 2018-10-23
Remove path as global option

rylandg:
Removed global `path` option and only retain it as a specific option for `create` and `deploy`.


rylandg:
Not at all.

PR #361, opened 2018-10-30
Imagemagick tested for bolt addition

rylandg:
This spec test verifies that our bolt runtime has a valid and (relatively recent version) of imagemagick installed.

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -206,6 +206,7 @@
       - bn remove internalnets
   steps:
       - in: bn invoke internalnets
+
 - test: Test logs (good-path)
   setup:
     - export FUNC_NAME=rand$(cat /dev/urandom | tr -dc 'a-z0-9' | fold -w 32 | head -n 1)
@@ -719,6 +720,26 @@
     - in: bn invoke largeData200MB
       out: |-
           "Hello World!"
+
+- test: Test if imagemagick exists in bolt
+  setup:
+      - bn create node8 imagemagick
+      - npm i shelljs
+      - |-
+        cat <<EOF >/home/dockeruser/test/function.js
+        
+        const shell = require('shelljs');
+        exports.handler = async() => {
+          return shell.exec('convert --version');
+        };
+        EOF
+      - bn deploy imagemagick
+  cleanup:
+      - bn remove imagemagick
+  steps:
+      - in: bn invoke imagemagick
+        out: |-
+            "Version: ImageMagick 6.8*
 
 - test: Stats help
   steps:
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -721,7 +721,7 @@
       out: |-
           "Hello World!"
 
-- test: Test if imagemagick exists in bolt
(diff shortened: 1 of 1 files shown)


test/spec.yml
```diff
@@ -720,6 +721,26 @@
...
+- test: Test if imagemagick exists in bolt
+  setup:
+      - bn create node8 imagemagick
+      - npm i shelljs
+      - |-
+        cat <<EOF >/home/dockeruser/test/function.js
+        
+        const shell = require('shelljs');
+        exports.handler = async() => {
+          return shell.exec('convert --version');
```
vogre: Additional npm dependency can be avoided by using `child_process.execSync('convert --version', { econding: 'utf8' });`

test/spec.yml
```diff
@@ -720,6 +721,26 @@
...
+        };
+        EOF
+      - bn deploy imagemagick
+  cleanup:
```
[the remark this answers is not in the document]
rylandg: Setup is different because it's not the core functionality we are testing. I know it seems like a small difference but to me it's a reasonable logical separation.

test/spec.yml
rylandg: I considered it at the time but I consider `shelljs` to be cleaner. I'm of the belief that most users wouldn't use `execSync` they would favor the npm package.

test/spec.yml
```diff
@@ -720,6 +721,26 @@
       out: |-
           "Hello World!"
 
+- test: Test if imagemagick exists in bolt
```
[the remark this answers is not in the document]
rylandg: Oh I see, using `bolt` was silly. Will change.

test/spec.yml
vogre: The top stackoverflow comment does not mention `shelljs`, but some do - it seems like an overkill since `shelljs` is "a portable (Windows/Linux/OS X) implementation of Unix shell commands on top of the Node.js API".

However, this comment is specifically about an npm dependency to tests that are somewhat flaky.

rylandg: You may be right regarding usage (not sure). 

I definitely don't understand this comment

> However, this comment is specifically about an npm dependency to tests that are somewhat flaky.

The binaris.yml spec tests are only as flaky as the underlying system.

rylandg:
@michaeladda I changed the test name and removed the version check.

test/spec.yml
michaeladda: iiuc, the npm dependency adds flakiness in this case.

test/spec.yml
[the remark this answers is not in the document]
michaeladda: i didn't ask about setup but about cleanup. i expect that section to follow the steps, not the other way around. that's all.

test/spec.yml
vogre: There's flakiness of the underlying system (which we try to stress by running parallel tests), which is interesting (signal).

This adds the flakiness of `npm i` which is not related to the underlying system (noise).

Issue #166, opened 2017-11-15
inform the user about log levels when deployment fails


lavie:
e.g.
```
~/binaris $ cd user/
~/binaris/user $ bn deploy
bn: Failed to deploy function nappyneck
```

This is practicaly useless as an error message...
At the very minimum explain to the user how to increase the log level.

rylandg:
@lavie As a starting point what about https://github.com/binaris/binaris/commit/e1968e4de440a933f7c4b1e2bd72fabfc8fe8507. The text is negotiable (clearly...) but I'm more interested in the expected user interaction.

lavie:
Good stuff.

michaeladda:
The debug tip now shows everywhere. A bit too pervasive imho, but you can no longer miss the log levels.

PR #364, opened 2018-10-31
Feature error tips

rylandg:
Initial PR to get feedback on the exact wording. Brought about because of this old issue.

https://github.com/binaris/binaris/issues/166

Tests failure seems unrelated.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -48,7 +48,11 @@ const exceptionWrapper = function tryCatchWrapper(funcToWrap) {
       await funcToWrap(options);
       process.exit(0);
     } catch (err) {
-      logger.error(err.message);
+      let message = err.message;
+      if (process.env.BINARIS_LOG_LEVEL === 'info' || !process.env.BINARIS_LOG_LEVEL) {
+        message = `${message}\nTip: You can export BINARIS_LOG_LEVEL=[silly|debug|verbose|info|warn|error] to view debug logs`;
+      }
+      logger.error(message);
       process.exit(1);
     }
   };
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,7 +51,7 @@ const exceptionWrapper = function tryCatchWrapper(funcToWrap) {
       logger.error(err.message);
       if (!process.env.BINARIS_LOG_LEVEL) {
         // eslint-disable-next-line max-len
-        logger.warn('\nTip: You can export BINARIS_LOG_LEVEL=[silly|debug|verbose|info|warn|error] to view debug logs');
+        logger.warn('\nTip: You can export BINARIS_LOG_LEVEL=[silly|debug|verbose] to view debug logs');
       }
       process.exit(1);
     }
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -47,6 +47,10 @@ const exceptionWrapper = function tryCatchWrapper(funcToWrap) {
       process.exit(0);
     } catch (err) {
       logger.error(err.message);
+      if (process.env.BINARIS_LOG_LEVEL === 'info' || !process.env.BINARIS_LOG_LEVEL) {
```
arielshaqed: This is odd.  E.g. it logs if I set `BINARIS_LOG_LEVEL=info`, but _not_ if i set `BINARIS_LOG_LEVEL=warn` or `BINARIS_LOG_LEVEL=quux`.

You definitely want to log if `!process.env.BINARIS_LOG_LEVEL`.  And arguably you might also want to log this if `!logger.isDebugEnabled` (or somesuch; also this works in Winston v3 I think but I dunno about v2).

lib/index.js
```diff
@@ -47,6 +47,10 @@ const exceptionWrapper = function tryCatchWrapper(funcToWrap) {
...
+        // eslint-disable-next-line max-len
+        logger.warn('\nTip: You can export BINARIS_LOG_LEVEL=[silly|debug|verbose|info|warn|error] to view debug logs');
```
arielshaqed: Is it really true e.g. for `error`?

rylandg: I mean those are the logging levels we have defined. Of course error level works.

```
const colorLevels = {
  silly: 'white',
  debug: 'white',
  verbose: 'white',
  info: 'white',
  warn: 'yellow',
  error: 'red',
};
```

lib/index.js
rylandg: I don't understand. It's not disabling logging. It's just showing you a tip about logging levels if the user doesn't have a log level defined or if INFO is set.

lib/index.js
arielshaqed: Again, the code is easy to understand (yay!).  But I disagree with what it does.  The context in which I see the hint message is that I just had an error message.  And this is when you tell me I could "export `BINARIS_LOG_LEVEL=...error... to view debug logs`".  It's a good time to suggest `verbose` or `debug` or `silly` (even then, _only_ those that will actually give extra information!), but it's hardly a good time to tell me about levels that are more restrictive than I already have.

lib/index.js
rylandg: Ok that's more direct feedback. I don't mind removing the `info` case.

lib/index.js
rylandg: Understood.

rylandg:
@arielshaqed your turn again

test/cases.yml
```diff
@@ -10,13 +10,19 @@
   steps:
     -   in: bn deploy -p /home/dockeruser/test/trollycopter icydiceymolecule
         err: "ENOENT: no such file or directory, open '/home/dockeruser/test/trollycopter/binaris.yml'"
+        out: |-
+            
+            Tip: You can export BINARIS_LOG_LEVEL=[silly|debug|verbose] to view debug logs
```
arielshaqed: Consider not repeating this each time, and just putting a `*` in most of the outputs.

PR #366, opened 2018-11-01
support user custom http ctx

michaeladda:
matching change to binaris/bolt#101


michaeladda:
merging despite apparent no review, since it was already approved, and only dismissed due to force push after rebase (with no code changes at all)

Issue #368, opened 2018-11-04
Unknown parameters should trigger a warning


adamatan:
# What Happened

I have just wasted a few frustrating minutes for a silly:

    bn invoke createWebhook --date

instead of

    bn invoke createWebhook --data

typo.

# Desired Behaviour
Our CLI should either warn for unknown arguments (or fail completely, the same way `git` does) when facing unknown arguments.

rylandg:
Here https://github.com/binaris/binaris/pull/372

michaeladda:
already fixed

PR #372, opened 2018-11-05
Feature no more invalid args

rylandg:
No longer accept invalid args in the CLI. Tested with added spec.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -61,7 +61,8 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         alias: 'p',
         describe: 'Use directory dir. "create" will create this directory if needed.',
         type: 'string',
-      });
+      })
+      .strict();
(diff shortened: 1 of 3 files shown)


rylandg:
Fixes https://github.com/binaris/binaris/issues/368

cli.js
```diff
@@ -28,6 +28,7 @@ const handleCommand = async function handleCommand(options, specificHandler) {
     msgAndExit(`Invalid subcommand ${cmdSeq[1]} for command ${cmdSeq[0]}`, true);
   }
 
+  console.log(options);
```
vogre: Forgotten debug print?

rylandg:
@vogre embarassing mistake, thanks for catching it.

rylandg:
Will update.

rylandg:
Agreed. We need both but many can be tested via the spec format.

PR #375, opened 2018-11-11
Change invocation test regex to support a large response

### LINKED ISSUE #374: Binaris tests fail when output exceeds 4k (or 64k)
As seen in [ci](https://ci.binaris.io/blue/organizations/jenkins/gobinaris%2Fprecious/detail/master/5012/pipeline/223/).
Reproduced by a test that logs `'a'`*65535 and `'z'` with a test on output `a*z`.

arikmaor:
Should fix #374

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -653,7 +653,7 @@
         invokestats* function_duration_us* #* #-#-#T#:#:#.#Z* #-#-#T#:#:#.#Z*
         invokestats* function_invocations* #* #-#-#T#:#:#.#Z* #-#-#T#:#:#.#Z*
     - in: bn stats --since "3m" --json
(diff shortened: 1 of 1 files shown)


test/spec.yml
```diff
@@ -653,7 +653,7 @@
         invokestats* function_duration_us* #* #-#-#T#:#:#.#Z* #-#-#T#:#:#.#Z*
         invokestats* function_invocations* #* #-#-#T#:#:#.#Z* #-#-#T#:#:#.#Z*
     - in: bn stats --since "3m" --json
-      out: '{*"metrics":[*]*}'
+      out: '{*"metrics":[*'
```
arikmaor: note: this is **NOT** a regex
it's a glob

PR #380, opened 2018-11-22
allow time limit in perf command

michaeladda:
tested by CI, improves user control over run time, and helpful for our internal tests


test/spec.yml
```diff
@@ -816,3 +816,62 @@
         exit: 1
         err: "ENAMETOOLONG: name too long, open '/home/dockeruser/test/.binaris/binaris*1111.tgz'"
 
+- test: Test perf 1 second (good-path)
```
rylandg: A more trustworthy test would rely on `time` and not internal reporting.

michaeladda:
i don't see what would be gained from implementing upper bound

PR #389, opened 2018-12-02
Use V2 URLs

bergundy:
Implements the V2 spec (for KubeCon).
Should theoretically support pre account URLs when `accountId` is not found to avoid breaking current customers' integrations.
Depends on integration from other services with this branch.

CC: @michaeladda

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -261,22 +261,18 @@ const statsHandler = async function statsHandler(options) {
  */
 const loginHandler = async function loginHandler() {
   logger.info(
-`Please enter your Binaris account ID and API key to deploy and invoke functions.
-If you don't have an account ID or key, head over to https://binaris.com to request one`);
-  const { accountId, apiKey } = await inquirer.prompt([
-    {
-      type: 'text',
-      name: 'accountId',
-      message: 'Account ID:',
-    },
+`Please enter your Binaris API key to deploy and invoke functions.
+If you don't have a key, head over to https://binaris.com to request one`);
+  const { apiKey } = await inquirer.prompt([
     {
       type: 'password',
       name: 'apiKey',
       message: 'API Key:',
     },
   ]);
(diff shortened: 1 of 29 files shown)


test/spec.yml
```diff
@@ -63,6 +63,32 @@
                   bn logs foo --since 13hours
                   bn logs foo --since 9s*
 
+- test: Test old endpoints
+  cleanup:
+    - bn remove momentousgiants
```
michaeladda: please use a new function name, to avoid confusion.

test/spec.yml
```diff
@@ -63,6 +63,32 @@
...
+              (use "bn create node8 hello" to create a Node.js template function in your CWD)
+    -   in: bn deploy momentousgiants
+        out: |-
+          Deployed function momentousgiants
+          Invoke with one of:
+            "bn invoke momentousgiants"
+            "curl -H X-Binaris-Api-Key:* https://*"
+    -   in: bn invoke momentousgiants
+        out: '"Hello World!"'
+
```
michaeladda: please check `bn logs` and `bn list` as well for the old endpoints

lib/index.js
```diff
@@ -259,15 +263,30 @@ const loginHandler = async function loginHandler() {
...
+
+  const apiKeyIdx = rawApiKey.lastIndexOf('-');
+  const apiKey = rawApiKey.substr(apiKeyIdx + 1);
+  const realm = rawApiKey.substr(0, apiKeyIdx);
+
+  if (realm) {
+    forceRealm(realm);
   }
-  await updateAPIKey(answer.apiKey);
+  const { accountId, error } = await auth.verifyAPIKey(apiKey);
```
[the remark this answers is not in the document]
bergundy: I wanted to preserve the behaviour of the previous login flow

michaeladda:
please notice that since there are multiple contributors to this PR, i've switched to use merges and not rebases so we do not accidentally force push over each other changes.

PR #397, opened 2018-12-06
Allow underscores in function names

vogre:
Not changing the functionality of validation on `cli` to avoid the round-trip to binaris API.

### DIFF
--- a/lib/nameUtil.js
+++ b/lib/nameUtil.js
@@ -1,7 +1,7 @@
 'use strict';
 
 // allows only letters and numbers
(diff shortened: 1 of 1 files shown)


vogre:
Validation is really about an early warning, instead of only giving the warning after code upload is done - the backend is the ultimate authority on allowed names.

Tested locally (the backend was already accepting underscores).

For now no need to patch bump since this is merged into `feature-v2-run` which is only published as 6.0.0-rc.3.

PR #406, opened 2018-12-09
warn if using empty env

### LINKED ISSUE #369: Functions can be deployed with undefined secrets / env
# The Problem
Our CLI lets the user deploy a function even if the keys in its `env` clause are undefined in the local environment.

This happens a lot when I open a new terminal, deploy from it, and get strange errors from my running function:

    500 - "{\"error\":\"401 - \\\"invalid key\\\"\",\"stack\":\"StatusCodeError: 401 - \\\"invalid key\\\"\\n...

# Suggested solutions
1. We can prevent this annoyance by checking whether everything in the `env` clause is defined before deployment.
2. We can create an SDK for our secrets. This will be useful if future secrets will not be kept as env vars. When the code calls `secrets.get`, the function fails with `undefined secret`.

michaeladda:
fixes #369 
tested by CI

### DIFF
--- a/sdk/deploy.js
+++ b/sdk/deploy.js
@@ -74,8 +74,10 @@ const deployConf = async function deployConf(accountId, funcName, apiKey, funcCo
       }
       if (env[key] == null) {
         env[key] = process.env[key];
-        if (!env[key] || env[key] === '') {
+        if (env[key] === undefined) {
           logger.warn(`Ignoring empty env ${key}`);
+        }
+        if (!env[key] || env[key] === '') {
           delete env[key];
         }
       } else if (typeof env[key] !== 'string') {
--- a/sdk/deploy.js
+++ b/sdk/deploy.js
@@ -75,9 +75,12 @@ const deployConf = async function deployConf(accountId, funcName, apiKey, funcCo
       if (env[key] == null) {
         env[key] = process.env[key];
         if (env[key] === undefined) {
-          logger.warn(`Ignoring empty env ${key}`);
+          logger.warn(`Ignoring non existing env var '${key}'`);
         }
-        if (!env[key] || env[key] === '') {
+        if (env[key] === '') {
+          throw new Error(`Empty existing env var '${key}' is not supported`);
+        }
+        if (!env[key]) {
           delete env[key];
         }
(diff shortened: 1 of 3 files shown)


sdk/deploy.js
```diff
@@ -75,6 +75,7 @@ const deployConf = async function deployConf(accountId, funcName, apiKey, funcCo
       if (env[key] == null) {
         env[key] = process.env[key];
         if (!env[key] || env[key] === '') {
+          logger.warn(`Ignoring empty env ${key}`);
```
arielshaqed: Then how do I pass an empty value?

I'd understand warning `if (! key in process.env)`, but I *should be* allowed explicitly to specify an empty value for a variable.

michaeladda:
@arielshaqed PTAL

sdk/deploy.js
```diff
@@ -74,8 +74,10 @@ const deployConf = async function deployConf(accountId, funcName, apiKey, funcCo
       }
       if (env[key] == null) {
         env[key] = process.env[key];
-        if (!env[key] || env[key] === '') {
+        if (env[key] === undefined) {
           logger.warn(`Ignoring empty env ${key}`);
+        }
+        if (!env[key] || env[key] === '') {
```
arielshaqed: Nit: `''` is falsy, so `if (!env[key])` is enough.

But also I'm missing something: this again prevents me from sending an empty environment variable.  (The test below seems to say the same.)

sdk/deploy.js
arielshaqed: Thanks!

test/spec.yml
```diff
@@ -1054,3 +1055,21 @@
...
+  steps:
+    -   in: bn create node8 emptyenv
+    -   in: |-
+            cat <<EOF >>/home/dockeruser/test/binaris.yml
+                env:
+                    nosuchenv:
+            EOF
+    -   in: bn deploy emptyenv
+        out: |-
+            Ignoring non existing env var 'nosuchenv'*
```
arielshaqed: Nit: use `%` here (restricts to single line, added in #398).

michaeladda: unfortunately, there are multiple lines there (all the deploy blurb)

PR #408, opened 2018-12-09
 print errors in binaris.yml

michaeladda:
tested with CI

### DIFF
--- a/lib/binarisYML.js
+++ b/lib/binarisYML.js
@@ -30,10 +30,16 @@ const saveBinarisConf = async function saveBinarisConf(funcDirPath, binarisConf)
 
 const getFunctionsSection = function getFunctionsSection(binarisConf) {
   // ensure configuration has the expected field
-  if (!Object.prototype.hasOwnProperty.call(binarisConf, funcStr)) {
+  if (typeof binarisConf !== 'object' ||
+      binarisConf === null ||
+      !Object.prototype.hasOwnProperty.call(binarisConf, funcStr)) {
     throw new Error(`Your ${binarisConfFile} did not contain a require field: <${funcStr}>`);
   }
   const funcSection = binarisConf[funcStr];
+  if (typeof funcSection !== 'object' ||
+      funcSection === null) {
+    throw new Error(`Your ${binarisConfFile} must contain a <${funcStr}> dictionary.`);
+  }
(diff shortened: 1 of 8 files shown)


lib/binarisYML.js
```diff
@@ -30,10 +30,16 @@ const saveBinarisConf = async function saveBinarisConf(funcDirPath, binarisConf)
 
 const getFunctionsSection = function getFunctionsSection(binarisConf) {
   // ensure configuration has the expected field
-  if (!Object.prototype.hasOwnProperty.call(binarisConf, funcStr)) {
+  if (typeof binarisConf !== 'object' ||
+      binarisConf === null ||
+      !Object.prototype.hasOwnProperty.call(binarisConf, funcStr)) {
     throw new Error(`Your ${binarisConfFile} did not contain a require field: <${funcStr}>`);
```
arielshaqed: nit: "require*d*".

For that matter, why are we logging errors in the past tense, with pronouns?  I think I'm more used to seeing e.g. "`${binarisConfFile} is missing required field <${funcStr}>`" or similar.  (But these can maybe wait for a separate PR.)

lib/binarisYML.js
```diff
@@ -30,10 +30,16 @@ const saveBinarisConf = async function saveBinarisConf(funcDirPath, binarisConf)
...
   }
   const funcSection = binarisConf[funcStr];
+  if (typeof funcSection !== 'object' ||
+      funcSection === null) {
+    throw new Error(`Your ${binarisConfFile} must contain a <${funcStr}> dictionary.`);
```
arielshaqed: Again, this message is *really* confusing.

sdk/deploy.js
```diff
@@ -60,35 +60,25 @@ const deployConf = async function deployConf(accountId, funcName, apiKey, funcCo
   const { env } = funcConf;
   if (env) {
     if (typeof env !== 'object') {
-      throw new Error('binaris.yml env section must be a dictionary.');
+      throw new Error('binaris.yml section <env> is not a dictionary');
```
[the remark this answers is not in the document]
michaeladda: Why? funcConf was built from userConfPath.
Or maybe I'm missing something.
Do you want to print instead of binaris.yml the full path?

PR #413, opened 2018-12-11
use only package version as git tag

michaeladda:
since npm version are immutable, but npm tags are mutable,
it makes more sense to have the git tag match the immutable npm
specifier.

not tested unfortunately. will test soon, when i'll publish a new cli


michaeladda:
Discussed the jenkins jobs and approved face to face with @arielshaqed. merging.

PR #416, opened 2018-12-11
Allow only ASCII characters in API key and account ID fields

arielshaqed:
They must be compatible with HTTP headers, and we don't bother to encode them according to the RFCs.  So they should be just ISO 8859-1, and because we don't care about those extra characters we just go with ASCII.

Why not encode non-ASCII into the headers?  Apart from being a lot of work for no gain, there's actual harm in having to support multiple encodings of the same field.  E.g. much fun would be had with X.509 / PKCS certificates had DNs not been encoded canonically.  ASCII is so much easier to think about, and we control both ends to can do as we like.

### DIFF
--- a/lib/userConf.js
+++ b/lib/userConf.js
@@ -69,11 +69,35 @@ const getConf = async function getConf(key, envVar, ...defaultValue) {
   return userConf[key];
 };
 
+// Matches printable ASCII-only values.
+const ascii = /^[\u0020-\u007E]*$/;
+
+// Validates a header can be sent.  In practice ISO 8859-1 is allowed,
+// but that is rejected by the Binaris backend.  Disallow it on
+// principle, for aesthetic reasons
+const validateHeaderValue = function validateHeaderValue(value, context) {
+  if (!ascii.test(value)) {
+    throw new Error(`Non-ASCII characters are not allowed in ${context}.`);
+  }
+  return value;
+};
+
+// Wraps func in validateHeaderValue.
+const validateGet = function validateGet(func, context) {
+  return async function get(...args) { return validateHeaderValue(await func(...args), context); };
+};
+
+// Wraps func argument in validateHeaderValue.
+const validateUpdate = function validateUpdate(func, context) {
+  return async function update(value) { await func(validateHeaderValue(value, context)); };
+};
+
 module.exports = {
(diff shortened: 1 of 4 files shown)


test/spec.yml
```diff
@@ -69,6 +69,7 @@
                   bn logs foo --since 9s*
 
 - test: Test old endpoints
+  serial: true
```
michaeladda: nice catch!

lib/userConf.js
```diff
@@ -69,11 +69,34 @@ const getConf = async function getConf(key, envVar, ...defaultValue) {
...
 
+// Matches printable ASCII-only values.
+const ascii = /^[\u0020-\u007E]*$/;
+
+// Validates a header can be sent.  In practice ISO 8859-1 is allowed,
+// but that is rejected by the Binaris backend.  Disallow it on
+// principle, for aesthetic reasons
+const validateHeaderValue = async function validateHeaderValue(value, context) {
+  if (!ascii.test(value)) {
+    throw new Error(`Non-ASCII characters are not allowed in ${context}.`);
```
michaeladda: afaict all other errors do not end in a comma.
i don't have strong preference either way, but i do strongly prefer consistency.
please make it consistent

lib/userConf.js
```diff
@@ -69,11 +69,34 @@ const getConf = async function getConf(key, envVar, ...defaultValue) {
   return userConf[key];
 };
 
+// Matches printable ASCII-only values.
+const ascii = /^[\u0020-\u007E]*$/;
```
michaeladda: empty strings should probably be disallowed too

lib/userConf.js
arielshaqed: Thanks; done!

lib/userConf.js
arielshaqed: *Not* done.
1. This is _not_ general enforcement of having nice values, this is specific enforcement that these values can easily be transferred in HTTP headers without the request client complaining.
2. An empty string is a perfectly good string.  Why enforce it on the client?

arielshaqed:
@michaeladda , PTAL!

arielshaqed:
Version bumped.  @michaeladda can I pull now?

PR #418, opened 2018-12-15
Update default endpoints

rylandg:
Needed for https://github.com/binaris/functions-examples/pull/20

Tested with spec


rylandg:
Responded to the issue.

rylandg:
After discussing in the issue linked I understand now that there isn't a real problem fixed by this code. I will instead release a patch unifying internal endpoints and output endpoints.

PR #425, opened 2019-01-07
Feature use cleanup in tests

rylandg:
Pretty straightforward. When using remove outside of the cleanup step there is a chance for dangling state.

Who likes dangling state?


rylandg:
Not as thoroughly as I should have. I mostly just noticed with the `js` CURL tests when I started writing python curl tests. Will go over it.

PR #429, opened 2019-01-14
Include debug ID in internal error messages

vogre:
Re-ran multiple times to get the error to show up.

Done for specific `ERR_INTERNAL` and not in a generic way since extra info is probably unneeded for errors like `No such function` or `Invalid API key`.


vogre:
Closing in favour of current server-side solution.

PR #430, opened 2019-01-29
Add concurrent execution model test

arikmaor:
Should be merged right after https://github.com/binaris/marathon/pull/149

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -457,6 +457,48 @@
         exit: 1
         err: "Empty string env 'EMPTY' in binaris.yml is not supported"
 
+- test: "concurrent execution model: deploy invoke multiple times"
+  steps:
+    -   in: |-
+            cat <<EOF >/home/dockeruser/test/binaris.yml
+            functions:
+              concurrentfunc:
+                file: function.js
+                entrypoint: handler
+                runtime: node8
+                executionModel: concurrent
+            EOF
+    -   in: |-
+            cat <<EOF >/home/dockeruser/test/function.js
+            function msleep(ms) {
+              return new Promise((resolve) => {
+                  setTimeout(resolve, ms);
+              });
+            }
+            let duringFlight = false;
+            exports.handler = async () => {
+              if (duringFlight) {
+                return true;
+              } else {
+                duringFlight = true;
+                await msleep(10000);
+                duringFlight = false;
+                return false;
+              }
+            }
+            EOF
+    -   in: bn deploy concurrentfunc
+        out: |-
+            Deployed function concurrentfunc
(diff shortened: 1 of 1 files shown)


test/spec.yml
```diff
@@ -457,6 +457,47 @@
         exit: 1
         err: "Empty string env 'EMPTY' in binaris.yml is not supported"
 
+- test: "concurrent execution model: deploy invoke multiple times"
+  steps:
+    -   in: |-
+            cat <<EOF >/home/dockeruser/test/binaris.yml
```
rylandg: I'm not sure why we started writing tests with the full absolute path but these tests run from the `test` directory and do not need it. Is there an advantage to this method?

bergundy: Good point @rylandg

test/spec.yml
```diff
@@ -457,6 +457,47 @@
...
+              }
+            }
+            EOF
+    -   in: bn deploy concurrentfunc
+        out: |-
+            Deployed function concurrentfunc
+            Invoke with one of:
+              "bn invoke concurrentfunc"
+              "curl -H X-Binaris-Api-Key:* https://*"
+    -   in: "(for i in `seq 1 12`; do bn invoke concurrentfunc & done) | grep true"
```
bergundy: Why 12?

test/spec.yml
```diff
@@ -457,6 +457,47 @@
...
+                executionModel: concurrent
+            EOF
+    -   in: |-
+            cat <<EOF >/home/dockeruser/test/function.js
+            function msleep(ms) {
+              return new Promise((resolve) => {
+                  setTimeout(resolve, ms);
+              });
+            }
+            let duringFlight = false;
```
bergundy: Maybe sleep always and hold a counter of inflight requests?
Would be nice to know how much concurrency we actually achieve

test/spec.yml
```diff
@@ -884,7 +925,7 @@
       - npm i shelljs
       - |-
         cat <<EOF >/home/dockeruser/test/function.js
-        
+
```
bergundy: It's better to separate white space fixes and non related changes from PRs that change or add functionality.

test/spec.yml
arikmaor: we were a bit tired when deciding on 12, we actually need `minBolts+1` so in the `testing` realm, that would be 3

test/spec.yml
arikmaor: it would be much difficult to test, now I can just do ` | grep true`

test/spec.yml
arikmaor: agree

PR #433, opened 2019-01-30
Feature execution model option

rylandg:
This adds the ability to specify execution model at create time. Additionally it sets a default model of "exclusive"

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -42,6 +42,8 @@ const handleCommand = async function handleCommand(options, specificHandler) {
 
 const runtimes = require('./lib/runtimes');
 
+const executionModels = ['exclusive', 'concurrent'];
+
 const pathOption = ['path', {
   alias: 'p',
   describe: 'Use directory dir.',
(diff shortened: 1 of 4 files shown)


cli.js
```diff
@@ -42,6 +42,8 @@ const handleCommand = async function handleCommand(options, specificHandler) {
 
 const runtimes = require('./lib/runtimes');
 
+const executionModels = ['exclusive', 'concurrent'];
```
michaeladda: this should be derived from the conf type

cli.js
```diff
@@ -65,6 +67,12 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         describe: 'Function name',
         type: 'string',
       })
+      .option('executionModel', {
```
michaeladda: I don't see why this is required during create.
We should add this to the template function if it's not there.

cli.js
rylandg: This should be derived from the what?

cli.js
rylandg: Because it's a value configured in the yml. The only two comparable features are 

1. Runtime selection (forcefully required at `create` time)
2. Secrets, not written by the template and not required by the user ever

cli.js
```diff
@@ -65,6 +67,13 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        alias: 'e',
+        choices: executionModels,
+        describe: 'Execution model for your function',
+        default: 'exclusive',
```
michaeladda: the default should not be determined here. if the flag is missing just don't alter the templated yml. in the future this will be automatically detected unless specified, so no specification is a strong and important signal from the user (iiuc, using undefined as a default should be an easy solution)

PR #435, opened 2019-02-04
Add option to override HTTP method.

arielshaqed:
@arikmaor FYI this may help you invoke your backend functions manually.


arielshaqed:
Note there is no simple way to return the response body on a `HEAD` request.  Given that it seems so unuseful  that `request-promise` went out of their way to block returning its contents, I propose to take no action to return it.

vogre:
@arielshaqed: perhaps `HEAD` request behaviour should be similar to `curl -I`, print out the headers.

PR #436, opened 2019-02-05
Add "show" command: show configured API key and/or account ID

arielshaqed:


cli.js
```diff
@@ -284,6 +284,23 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        describe: 'Show configured account ID',
+        requiresArg: false,
+      })
+      .option('api_key', {
+        describe: 'Show configured account API key',
+        requiresArg: false,
+      })
+      .check(
+        (argv) => {
+          if (!argv.account_id && !argv.api_key) throw new Error('Must specify what to show.');
```
rylandg: This can be done in a much cleaner way (I believe) using the `demand` option. You can simply use `demand: 1` which will force 1 arg.

test/spec.yml
```diff
@@ -124,6 +125,13 @@
             "bn invoke momentousgiants"
             "curl -H X-Binaris-Api-Key:* https://*"
 
+- test: Test show
+  steps:
+    - in: BINARIS_ACCOUNT_ID=1066 bn show --account_id
```
rylandg: I would strongly prefer that a file based test also exists.

lib/index.js
```diff
@@ -268,6 +268,14 @@ const statsHandler = async function statsHandler(options) {
   }
 };
 
+/**
+ * Show information about configured account.
+ */
+const showHandler = async function showHandler(options) {
+  if (options.account_id) logger.info(await getAccountId());
```
rylandg: What happens in case of failures?

arielshaqed:
```sh
flutterby:binaris ariels$ chmod -r ~/.binaris.yml
flutterby:binaris ariels$ node ./cli.js show --apiKey  --accountId
Binaris conf file could not be read and BINARIS_ACCOUNT_ID is undefined, please use "bn login"
```

test/spec.yml
arielshaqed: Not done.  I'm not testing that `getAccountId` works here -- if it doesn't, _nothing_ will work.

arielshaqed:
@rylandg: I made (most) requested changes; PTAL.

test/spec.yml
```diff
@@ -124,6 +125,13 @@
             "bn invoke momentousgiants"
             "curl -H X-Binaris-Api-Key:* https://*"
 
+- test: Test show
+  steps:
+    - in: BINARIS_ACCOUNT_ID=1066 bn show --accountId
```
vogre: What do you think about using a positional, for `bn show apiKey`, a bit like `git config user.name`?

arielshaqed: I wanted to do it initially, but then you get into weirdnesses about supporting multiple positionals.  @rylandg WDYT?

rylandg: I think multiple positional arguments is pretty easy with `yargs`. I personally am indifferent in regards to the decision.

arielshaqed:
Added positional args, pushing once test pass.

Thanks!

PR #439, opened 2019-02-17
Add definitions and autocompletions to js template function

arikmaor:


lib/functionTemplates/js/binaris.d.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+declare class BinarisHttpResponse {
+  userResponse: Partial<FunctionResponse>;
+  constructor(userResponse: Partial<FunctionResponse>);
+}
+
+declare type FunctionContext = {
+  request: BinarisHttpRequest;
+  HTTPResponse: typeof BinarisHttpResponse;
```
michaeladda: HTTP and Http. I don't have a preference, but I like consistency

lib/functionTemplates/js/binaris.d.ts
```diff
@@ -0,0 +1,28 @@
+/// <reference types="node" />
+interface BinarisHttpRequest {
+  env: Record<string, string | undefined>;
+  query: Record<string, string | string[] | undefined>;
+  body: Buffer;
+  path: string;
+  method: string;
```
michaeladda: this can be more restrictive ( 'GET' | 'PUT' | ...) but I'm not sure it's valuable

rylandg: There are some non-standard headers which I assume we allow users to support via custom `HTTPResponse` such as `PATCH`.

lib/functionTemplates/js/binaris.d.ts
```diff
@@ -0,0 +1,28 @@
...
+  userResponse: Partial<FunctionResponse>;
+  constructor(userResponse: Partial<FunctionResponse>);
+}
+
+declare type FunctionContext = {
+  request: BinarisHTTPRequest;
+  HTTPResponse: typeof BinarisHTTPResponse;
+};
+
+export declare type BinarisFunction = (body: unknown, context: FunctionContext) => Promise<any>
```
rylandg: Don't we know the generic form of the body?

arikmaor: It can be whatever you pass in the body...

PR #440, opened 2019-02-18
Change node default execution model to concurrent

bergundy:


lib/functionTemplates/java8/binaris.yml
```diff
@@ -3,3 +3,4 @@ functions:
         file: build/libs/example.jar
         entrypoint: ExampleFunction
         runtime: java8
+        executionModel: exclusive
```
arikmaor: this is the default, why do we need to specify it?

bergundy:
@arielshaqed 
1. The default in spice is just for backwards compatiblity, I'll change the default there to be runtime dependant once this is approved.
2. I don't see why this is a problem, I suspect python code would be mostly synchronous while nodejs would be async by the nature of these languages.

lib/functionTemplates/java8/binaris.yml
rylandg: I think it add clarity.

bergundy:
@ryland do you think that node should default to exclusive?

arikmaor:
> I'm not sure how much it actually impacts users but I personally see different defaults as confusing

I agree, we should decide one default for all

lib/functionTemplates/java8/binaris.yml
arikmaor: what do you understand when you look at it? (pretend that you only know java)

rylandg: I know that I can google `executionModel: exclusive` and understand what is going on (and that there are other options).

bergundy:
I'm afraid that if we don't default to `concurrent` in node we'd be in trouble since we'll never be able to change it once we have users relying on this behavior.

@michaeladda @avnerbraverman please make a call here.

bergundy:
@arikmaor we should probably add a comment about what `executionModel` is in the template yml perhaps with a link describing the feature.
I'd rather expose this to the users so they know how their function runs and that another option exists.

arielshaqed:
> @michaeladda @avnerbraverman please make a call here.

Not sure I agree with the solution suggested by @bergundy, but we *do* need a resolution here... and it has far-reaching consequences.

lib/functionTemplates/java8/binaris.yml
arikmaor: are you sure you can google it?

bergundy:
Ping @michaeladda @avnerbraverman

michaeladda:
Sorry, I was sure I commented before.
1. I don't find different default model per runtime too confusing.
1. I would like to have shared as default for nodejs
1. I don't want to make a breaking changes (=avoiding surprises when a user re-deploys a function)
Therefore, I think a good way to shift the users to shared execution model is to bundle it with a new feature, such as node10 runtime for example, and then deprecating the old runtime and the old default execution model

bergundy:
@michaeladda I don't really see the risk here

bergundy:
If nobody can find a good reason why this is a breaking change, I'm merging this PR.

arielshaqed:
On Thu, 28 Feb 2019, 16:36 Roey Berman, <notifications@github.com> wrote:

> If nobody can find a good reason why this is a breaking change, I'm
> merging this PR.
>
I asked @michaeladda to call it, and he did. Redeploying will be confusing
but doesn't break anything afaict. Do merge.

>

Issue #428, opened 2019-01-14
debug ids should be printed on errors


michaeladda:
currently `validateResponse` in `binaris/sdk/handleError.js` does not print `debugId`. Some errors may have it in their body, but some don't and hiding the debug id from the user prevents her from being able to get more info via inquiry

michaeladda:
To clarify, `debugId` in this context is the already existing `X-Binaris-Request-ID` as returned from the backend. No need to give it a new name.

PR #442, opened 2019-02-20
Feature debug id in errors

rylandg:
I expect to get some strong pushback regarding the way I inject the debug ID string into the final output. I initially did so without modifying the `ApiError` constructor but this seemed to be much less DRY. 

Invoke was originally intended to be included with this patch but after doing some digging it appears that `bolt` does not even send a `debugID` with errors. I'm relatively sure this is an oversight but if not I would love to understand the rationale. 

I tested this locally and via the spec tests (although they are meager).

### DIFF
--- a/sdk/handleError.js
+++ b/sdk/handleError.js
@@ -11,19 +11,20 @@ const { maybeTranslateErrorCode } = require('binaris-pickle');
 const { version } = require('../package.json');
 
 class APIError extends Error {
-  constructor(message, requestIdString) {
-    super(`${requestIdString || ''}${message}`);
+  constructor(message, requestId) {
+    super(message);
+    const requestIdString = requestId ? `RequestId: ${requestId}\n` : '';
+    this.message = `${requestIdString}${message}`);
   }
 }
 
 function validateResponse(response) {
   const requestId = get(response, 'headers.x-binaris-request-id');
-  const requestIdString = requestId ? `RequestId: ${requestId}\n` : '';
   const errorCode = get(response, 'body.errorCode');
   if (errorCode) {
     const readableErrorMessage = maybeTranslateErrorCode(errorCode);
     if (readableErrorMessage) {
-      throw new APIError(`Error: ${readableErrorMessage}`, requestIdString);
+      throw new APIError(`Error: ${readableErrorMessage}`, requestId);
     }
   }
 
@@ -31,15 +32,15 @@ function validateResponse(response) {
   // with an untranslated message.
   const errorText = get(response, 'body.error') || (errorCode && get(response, 'body.message'));
   if (errorText) {
-    throw new APIError(errorText, requestIdString);
+    throw new APIError(errorText, requestId);
   }
 
   if (errorCode) {
-    throw new APIError(`Error: ${errorCode}`, requestIdString);
+    throw new APIError(`Error: ${errorCode}`, requestId);
   }
 
   if (response.statusCode < 200 || response.statusCode >= 300) {
-    throw new APIError(`Unexpected response ${response.statusCode}`, requestIdString);
+    throw new APIError(`Unexpected response ${response.statusCode}`, requestId);
   }
 
   return response;
--- a/sdk/handleError.js
+++ b/sdk/handleError.js
@@ -10,14 +10,20 @@ const { maybeTranslateErrorCode } = require('binaris-pickle');
 
 const { version } = require('../package.json');
 
-class APIError extends Error {}
+class APIError extends Error {
+  constructor(message, debugIDString) {
+    super(`${debugIDString || ''}${message}`);
+  }
+}
 
 function validateResponse(response) {
+  const debugId = get(response, 'body.debugId');
+  const debugIdString = debugId ? `DebugId: ${debugId}\n` : '';
   const errorCode = get(response, 'body.errorCode');
   if (errorCode) {
     const readableErrorMessage = maybeTranslateErrorCode(errorCode);
     if (readableErrorMessage) {
-      throw new APIError(`Error: ${readableErrorMessage}`);
+      throw new APIError(`Error: ${readableErrorMessage}`, debugIdString);
     }
   }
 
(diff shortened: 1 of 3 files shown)


test/spec.yml
```diff
@@ -864,10 +868,12 @@
     -   in: bn create node8 perplexingpersimmon
     -   in: bn deploy perplexingpersimmon
         err: |-
+            DebugId: *
```
vogre: I am not strongly against it, but the issue I see with adding `DebugId` to every error coming from the system is that most errors are user errors which we can't provide support for: if users have a typo in their key we can't know what the typo was, if user code size is 300MB we can't tell them what can be deleted, etc.

rylandg:
@vogre Thanks for the great feedback

As for the general usefulness of the feature, I'm less opinionated here. The origin of the feature is this ticket opened by Michael https://github.com/binaris/binaris/issues/428. To me, the rationale seems pretty sound, but I'd love for him to chime in @michaeladda.

As for the `bn logs` suggestion, I think it's a great idea and would probably synergize well with this. With that in mind we still need to decide how much filtering we want to be responsible for in Picard.

michaeladda:
For me `X-Binaris-Request-ID` is the proper name of the debug ID. We should certainly only use one name and not confuse users.
I think that since I created #428 @vogre added `X-Binaris-Request-ID` to the JSON itself as well, so now the users can see it either in the header or in the body.

If there is already a unique identifier in the body, no point in adding it again.
If a unique identifier is missing, we need one in a format the user can see when using `bn invoke`

michaeladda:
(sorry, closed by mistake. re-opened)

rylandg:
Ok. So my current plan is to always present the id to the user as `RequestID`. Luckily 90% of this patch stays valid (minus some name changes). I will also add support for `RequestID` in `invoke` and then dismiss the current review.

sdk/invoke.js
```diff
@@ -33,24 +34,24 @@ const invoke = async function invoke(accountId, funcName, apiKey, funcData) {
...
+      const fullError = [`RequestId: ${requestId}`, parsedError.error || parsedError.message];
+
+      if (parsedError.stack) {
+        fullError.push(parsedError.stack);
       }
+      throw new Error(fullError.join('\n'));
     }
 
-    throw new Error(err.error || err);
+    throw new Error(fallbackError);
```
michaeladda: if err is not instanceof StatusCodeError how can err.error be meaningful? shouldn't this line be `throw err`?

sdk/invoke.js
```diff
@@ -33,24 +34,24 @@ const invoke = async function invoke(accountId, funcName, apiKey, funcData) {
...
-    } catch (nestedErr) {
-      logger.debug('Failed to parse JSON response', { nestedErr });
-    }
-
-    if (parsedError) {
-      const errorOrMsg = parsedError.error || parsedError.message;
-      if (parsedError.stack && errorOrMsg) {
-        throw new Error(`${errorOrMsg}\n${parsedError.stack}`);
+    const fallbackError = err.error || err;
+    if (err instanceof StatusCodeError) {
```
michaeladda: is this important?
what happens if another error is returned?
shouldn't the code try to parse anyhow?

sdk/handleError.js
```diff
@@ -10,30 +10,36 @@ const { maybeTranslateErrorCode } = require('binaris-pickle');
...
-class APIError extends Error {}
+class APIError extends Error {
+  constructor(message, requestIdString) {
+    super(`${requestIdString || ''}${message}`);
+  }
+}
 
 function validateResponse(response) {
+  const requestId = get(response, 'headers.x-binaris-request-id');
+  const requestIdString = requestId ? `RequestId: ${requestId}\n` : '';
```
michaeladda: Why do this here and not in the APIError constructor? Just pass to APIError requestId

sdk/invoke.js
```diff
@@ -33,24 +34,24 @@ const invoke = async function invoke(accountId, funcName, apiKey, funcData) {
...
+      const requestId = err.response.headers['x-binaris-request-id'];
+      let parsedError;
+      try {
+        parsedError = JSON.parse(err.error);
+      } catch (parseError) {
```
michaeladda: doesn't `JSON.parse(err.error)` only make sense if `err.error` exists?
if err.error is missing, there will always be a parse error and then parsedError.error will be the fallbackError which is really err which is a very convoluted way to write err :-/

I think the entire section can be way more readable.
##
```javascript
...
  } catch (err) {
    const errorMessage = [];
    const requestId = err.response.headers['x-binaris-request-id'];
    if (requestId) {
      errorMessage.push = `RequestId: ${requestId}`;
    }
    if (err instanceof StatusCodeError) { // maybe not needed
      if (err.error) {
        try {
          const subError = JSON.parse(err.error);
          errorMessage.push(subError.error || subError.message);
          if (subError.stack) {
            errorMessage.push(subError.stack);
          }
        } catch (_) {
           errorMessage.push(err.error);
        }
      } else {
        errorMessage.push(err.message);
      }
      throw new Error(errorMessage.join('\n'));
    }
    throw err;
  }
```
That said, I would refactor this entirely, to use APIError which is really ErrorWithRequestId,
and name the function handling this accordingly (something about handle nested error)

vogre: Note that `StatusCodeError.error` is supposedly deprecated (response.body instead?): https://github.com/request/promise-core/blob/master/lib/errors.js - it contains the response body - it might fail if it's `undefined` or if it's non-JSON (such as legacy Java responses).

sdk/handleError.js
```diff
@@ -10,30 +10,37 @@ const { maybeTranslateErrorCode } = require('binaris-pickle');
 
 const { version } = require('../package.json');
 
-class APIError extends Error {}
+class APIError extends Error {
+  constructor(message, requestId) {
+    super(message);
+    const requestIdString = requestId ? `RequestId: ${requestId}\n` : '';
+    this.message = `${requestIdString}${message}`;
```
michaeladda: why so complex?
```
const requestIdString = requestId ? `RequestId: ${requestId}\n` : '';
super(`${requestIdString}${message}`);
```
(since you're not using `this` before `super` it's ok plus you don't mutate `message`)

sdk/invoke.js
```diff
@@ -33,24 +34,21 @@ const invoke = async function invoke(accountId, funcName, apiKey, funcData) {
...
-    if (parsedError) {
-      const errorOrMsg = parsedError.error || parsedError.message;
-      if (parsedError.stack && errorOrMsg) {
-        throw new Error(`${errorOrMsg}\n${parsedError.stack}`);
-      }
-      if (parsedError.message && parsedError.request_id) {
-        throw new Error(`${parsedError.message}\nrequest_id: ${parsedError.request_id}`);
+    if (err instanceof StatusCodeError) {
+      const requestId = err.response.headers['x-binaris-request-id'];
+      const fullError = [`RequestId: ${requestId}`];
```
michaeladda: This can be undefined iiuc. Still ok imho

sdk/handleError.js
rylandg: Fixed.

sdk/invoke.js
rylandg: I don't think it can be (as long as our services always return it).

PR #444, opened 2019-03-01
Print everything when show has --all

### LINKED ISSUE #443: bn show should show everything with no params
Current response when running 'bn show' with not arguments is:

Show Binaris account configuration

Positionals:
  config  What to show  [required] [choices: "accountId", "apiKey"]

Show simply show everything when no arguments are provided.

rylandg:
Fix #443 (as requested by Avner)

This still rejects any args but `apiKey`, `accountId` or empty.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -280,9 +280,9 @@ const statsHandler = async function statsHandler(options) {
 const showHandler = async function showHandler({ config, all }) {
   const conf = await loadAndValidateUserConf();
   if (all) {
-    for (const confKey in conf) {
+    Object.keys(conf).forEach((confKey) => {
       logger.info(`${confKey}: ${conf[confKey]}`);
-    }
+    });
   } else {
     logger.info(conf[config]);
   }
--- a/lib/index.js
+++ b/lib/index.js
@@ -278,12 +278,21 @@ const statsHandler = async function statsHandler(options) {
  * Show information about configured account.
  */
 const showHandler = async function showHandler({ config, all }) {
+  const showMap = {};
   if (config === 'accountId' || all) {
-    logger.info(await getAccountId());
+    showMap.accountId = await getAccountId();
   }
   if (config === 'apiKey' || all) {
-    logger.info(await getAPIKey());
+    showMap.apiKey = await getAPIKey();
   }
+
+  Object.keys(showMap).forEach((showKey) => {
+    if (all) {
+      logger.info(`${showKey}: ${showMap[showKey]}`);
+    } else {
+      logger.info(`${showMap[showKey]}`);
+    }
+  });
 };
 
 /**
--- a/lib/index.js
+++ b/lib/index.js
@@ -278,6 +278,10 @@ const statsHandler = async function statsHandler(options) {
  * Show information about configured account.
  */
 const showHandler = async function showHandler({ config }) {
+  if (config === undefined) {
+    logger.info(`accountId: ${(await getAccountId())}`);
+    logger.info(`apiKey: ${(await getAPIKey())}`);
+  }
   if (config === 'accountId') logger.info(await getAccountId());
   if (config === 'apiKey') logger.info(await getAPIKey());
 };
--- a/lib/index.js
+++ b/lib/index.js
@@ -277,13 +277,13 @@ const statsHandler = async function statsHandler(options) {
 /**
  * Show information about configured account.
  */
-const showHandler = async function showHandler({ config }) {
-  if (config === undefined) {
-    logger.info(`accountId: ${(await getAccountId())}`);
-    logger.info(`apiKey: ${(await getAPIKey())}`);
+const showHandler = async function showHandler({ config, all }) {
+  if (config === 'accountId' || all) {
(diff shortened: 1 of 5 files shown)


vogre:
I agree that this seems somewhat less-user friendly, since when running `bn show` prints appropriate help it's possible to suggest additionaly options, like `--all` or `--json`.

michaeladda:
not sure if it's better to discuss the goal here or in #443 but since the responses are here, i'll add mine here too. @avnerbraverman FYI.
1. I think printing the entire configuration as `bn` sees it (after reading files, consuming env vars, etc) is valuable at least for debug and observability but possibly for many other use cases.
1. I agree the suggested behaviour inconsistent with the rest of the `bn show` interface and with `bn` in general. printing the field name feels awkward in that sense.
1. adding `--all` imho solves this nicely, but i'm not sure that the original issue is about printing all the values, but rather about the help output (which imho is good).
1. I don't think `--json` in place of `--all` since it's inconsistent as well. it should be an output modifier for the entire command (and the entire cli). so `bn show accountId --json` would print `"53356123"` for example (since it's a string) and `bn show --all --json` would print `{"accountId": "53356123","apiKey":"thisisasecret"}`

rylandg:
@michaeladda I've done my best to add the newly discussed `--all` command. There is unfortunately not an "out of the box" solution in yargs for allowing both of the following together

1. required positional arguments
2. an option which disables the need for said positional arguments

Instead I've made the positional arguments optional and done a manual check in `cli.js` to fail with the wrong input configuration. I did it this way so we can still print the yargs help output and add a custom informative error message.

LMKWYT

rylandg:
1. You can’t add flags as choices for a positional (dash is reserved so to
speak, first thing I tried. I thought of just “all” but after consulting
Avner I felt a flag felt better. I don’t mind changing it though.

2. How does having a special command fix the issue of headers? I’m the one
who originally added them (so clearly I get their value) and I’m not even
able to understand why going from (no-args->special command) would change
the importance of headers. Will our users memorize the output ordering
either way?

rylandg:
@michaeladda I made an attempt to achieve what was discussed on the call today. LMKWYT.

lib/index.js
```diff
@@ -277,9 +277,22 @@ const statsHandler = async function statsHandler(options) {
 /**
  * Show information about configured account.
  */
-const showHandler = async function showHandler({ config }) {
-  if (config === 'accountId') logger.info(await getAccountId());
-  if (config === 'apiKey') logger.info(await getAPIKey());
+const showHandler = async function showHandler({ config, all }) {
+  const showMap = {};
+  if (config === 'accountId' || all) {
```
michaeladda: sorry. this still have explicit field names in it. i think something along the lines of
```
// mock code
const conf = await loadUserConf();
if (all) {
  // sort?
  for field in conf {
    logger.info(`${conf[field]}`);
  }
} else {
   // check field, and then
   logger.info(`${field}: ${conf[field]}`);
}
```

rylandg: I understand now

arielshaqed:
Leaving this review to @michaeladda , who's more convinced of the customer value. :-)

rylandg:
@michaeladda I attempted to make the change you wanted. I think there may have been some implications to the suggested method you didn't take into account.

1. Not using explicit names means we have to load the conf using a generic call (you suggested `loadUserConf` but this would have meant a missing `~/.binaris.yml` doesn't cause a meaningful error). Instead I wrote the `loadAndValidateUserConf` which throws the error.

2. By using the generic call we change the actual behavior of existing `bn show`. This is because as of today `bn show apiKey` will attempt to use the environment variable `BINARIS_API_KEY` in addition to the conf file. This may be a unintended feature anyway but it definitely effects the behavior.

Lastly, the error in `getConf` moved into `loadAndValidateUserConf` and in the process lost a bit of verbosity (the error doesn't mention env anymore). I can spend more time and try and align these but I wanted to get your opinion first.

rylandg:
I will fix the help output.

As for `--json`, I don't understand this at all. 

> --json does not work (either support it or return an error for this command)

`--json` does not exist and is not referenced by the current `show` command. Am I missing something? Why would I add an error message for an option that doesn't exist? Under this rationale I should add an error for every possible flag a user might expect on any command.

michaeladda:
Indeed, after looking closely, it's not just `--json`. It seems the new command is not `strict`. Therefore running `bn show --all --foobar --json` just works. IIUC the following should fix this
```
diff --git a/cli.js b/cli.js
index 80ee5ec..b9aeac9 100755
--- a/cli.js
+++ b/cli.js
@@ -327,7 +327,8 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         alias: 'a',
         describe: 'Show it all',
         type: 'boolean',
-      });
+      })
+      .strict();
   }, async (argv) => {
     if (!argv.all && !argv.config) {
       msgAndExit('"bn show" requires positional arguments or the "--all" flag', true);
```
(and probably merits a test)

rylandg:
@michaeladda added the spec test

PR #445, opened 2019-03-03
Fix typescript interface

arikmaor:


michaeladda:
also, this highlights the fact the types here are not derived from the system itself, and are not tested either :-/

PR #448, opened 2019-03-05
Add create config option

michaeladda:


cli.js
```diff
@@ -78,8 +81,34 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+  // Create a function from node8 template with concurrent execution model
+  bn create node8 pirate --config.executionModel=concurrent
+
+  // Create a function from python2 template with exclusive execution model and FOO env
+  bn create python2 hello --config.executionModel=exclusive --config.env.FOO=bar
+`);
   }, async (argv) => {
+    // validating here and not in coerce to keep output consistent
+    if (argv.config !== undefined) {
+      const configType = typeof argv.config;
```
rylandg: 1. Not sure what you're protecting against here. No `type` is provided for the `config` option in `create`. This means that everything will come in as either an `object` or `boolean` (in the case that only a flag is passed `bn create node8 foo --config`). Unless I'm missing a parsing mechanism, I don't understand how it could ever be an array. 

2. I'm not 100% sure this is the correct way to check if something is an object. There are a lot of differing opinions [here](https://stackoverflow.com/questions/8511281/check-if-a-value-is-an-object-in-javascript) but it seems like the method you used might not cover `null`. Again, I don't think it matters because the only types that can reach that line of execution are `boolean` and `object`.

cli.js
```diff
@@ -78,8 +81,34 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+      if (configType !== 'object' || Array.isArray(argv.config)) {
+        msgAndExit(`Non object create configuration options: ${argv.config}`, true);
+      }
+      // transform boolean true  for --config.env.VAR to work properly
+      if (argv.config.env && typeof argv.config.env === 'object') {
```
rylandg: 1. This doesn't guard against `bn create python2 hello --config.env.FOO.BOO=bar`, it actually creates the broken `binaris.yml`.

test/spec.yml
```diff
@@ -1215,3 +1215,35 @@
...
+             testConfOpts:
+               file: function.js
+               entrypoint: handler
+               executionModel: exclusive
+               runtime: node8
+               env:
+                 PATH: null
+                 FOO: bar
+           EOF
+    -  in: diff binaris.yml expected.yml
```
rylandg: I don't think this actually tests anything (there are no output verification blocks).

test/spec.yml
```diff
@@ -1215,3 +1215,35 @@
...
+    -  in: |-
+           cat <<EOF >>/home/dockeruser/test/expected.yml
+           functions:
+             testConfOpts:
+               file: function.js
+               entrypoint: handler
+               executionModel: exclusive
+               runtime: node8
+               env:
+                 PATH: null
```
rylandg: Also not sure I like the literal null here (in comparison to empty arg)

lib/index.js
```diff
@@ -92,14 +95,19 @@ const gatherMeta = async function gatherMeta(options, forcePath = true) {
...
   await fse.mkdirp(funcPath);
   const funcName = await create(options.function, funcPath,
-    options.runtime, options.executionModel);
+    options.runtime, config);
   const optPath = options.path ? ` -p ${funcPath}` : '';
   const optName = options.function ? ` ${funcName}` : '';
+  const optConfig = verbose && config ? ` using ${inspect(config)}` : '';
   logger.info(
-`Created function ${funcName} in ${funcPath}
+`Created function ${funcName} in ${funcPath}${optConfig}
```
rylandg: Not sure I think the config should be printed (could go either way)

lib/create.js
```diff
@@ -40,10 +40,8 @@ const create = async function create(functionName, functionPath,
   const funcConf = {
     ...YMLUtil.getFuncConf(binarisConf, templateName),
     runtime,
+    ...config,
```
rylandg: I think this means someone could technically overwrite their runtime with config.

cli.js
michaeladda: 1. 1. as the tests verify, the code is protecting against `--config=bad` which will turn `argv.config` a string for example (so, it can be more than an object or a boolean)
   2. added to the test the case of `--config=aaa --config=bbb` which generates an array.
2. since yargs does not generate null values, we're safe.

cli.js
michaeladda: indeed. it doesn't do any value sanitization on purpose. it's simply a mechanism to modify the yaml at create time. a user can create an invalid configuration, just as is possible with an editor.

lib/create.js
michaeladda: yes, you are correct (this is good for concise testing code)

lib/index.js
michaeladda: it's not. `verbose` is `false`. look above. i started with it enabled and found it's too noisy, but decided to keep this extra "unused" code in case it's needed. if requested i'll happily delete it altogether.

test/spec.yml
michaeladda: null here is not a string but the value null in yaml. the generated yaml is legal and works as expected.
to actually have the value null you'd have to use `'null'`.
moreso, using our yaml package it's impossible to generate empty values.

test/spec.yml
michaeladda: it does. if it's non zero, the test fails. (as observed it during testing)

cli.js
```diff
@@ -78,8 +81,34 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         describe: 'Use directory dir. "create" will create this directory if needed.',
         type: 'string',
       })
-      .strict();
+      .strict()
+      .example(
+` // Create a function from python3 template
```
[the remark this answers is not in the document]
rylandg: 1. Sure
2. I prefer consistency more than anything so it was the right call.

cli.js
rylandg: 1. Didn't think of it, 100% agree
2. A comment explaining the rationale for the checks might be nice.

cli.js
rylandg: Hmm not sure I see eye to eye here. That being said, if its the behavior you were aiming for, it's not block worthy.

lib/index.js
rylandg: My bad, no problems.

rylandg: Although, sure why you wouldn't just `logger.verbose` with the config specifically?

test/spec.yml
rylandg: If `js-yaml` can't generate empty values then it's fine as is.

test/spec.yml
rylandg: Hmm, ok. My only comment is that this form of spec test is harder to discern at first glance (not specific to this PR but more of a general thought).

Issue #393, opened 2018-12-05
Constructed regex becomes slow on large inputs


vogre:
Consider using non-greedy matcher in:
https://github.com/binaris/binaris/blob/66df54ab4b6d5b63970820fea2c610bbd9f76efc/test/test.js#L100

PR #452, opened 2019-03-11
Support TTL for deployed functions

arikmaor:
This PR:
1. Adds 2 E2E tests for functions with TTL defined (that they are created and automatically removed)
2. Displays the expiration of functions (if exists) in `bn list` output

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -213,20 +213,33 @@ Latencies
 const listHandler = async function listHandler(options) {
   const listedFuncs = await list();
   if (options.json) {
-    const rawData = Object.keys(listedFuncs).map(key => ({
-      name: key,
-      lastDeployed: `[${listedFuncs[key].tags.latest.modifiedAt}]`,
-    }));
+    const rawData = Object.keys(listedFuncs).map((key) => {
+      const ret = {
+        name: key,
+        lastDeployed: `[${listedFuncs[key].tags.latest.modifiedAt}]`,
+      };
+      const expiration = listedFuncs[key].tags.latest.expiration;
+      if (expiration) {
+        ret.expiration = `[${expiration}]`;
+      }
+      return ret;
+    });
     logger.info(JSON.stringify(rawData));
   } else {
-    const aggr = {};
-    Object.keys(listedFuncs).forEach((key) => {
-      aggr[key] = listedFuncs[key].tags.latest.modifiedAt;
+    const aggr = Object.keys(listedFuncs).map((key) => {
+      const ret = {
+        function: key,
+        'last deployed': listedFuncs[key].tags.latest.modifiedAt,
+      };
+      const expiration = listedFuncs[key].tags.latest.expiration;
+      if (expiration) {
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -212,23 +212,29 @@ Latencies
  */
 const listHandler = async function listHandler(options) {
   const listedFuncs = await list();
+  const rawData = Object.keys(listedFuncs).map(key => ({
```
arielshaqed: Consider mapping over entries (AFAIR we do have `lodash` in here).

lib/index.js
```diff
@@ -212,23 +212,29 @@ Latencies
...
-      aggr[key] = listedFuncs[key].tags.latest.modifiedAt;
-    });
-    if (Object.keys(aggr).length > 0) {
-      logger.info(columnify(aggr, {
-        columns: ['Function', 'Last Deployed'],
-        config: { Function: { minWidth: 25 } },
-      }));
-    }
+  } else if (rawData.length > 0) {
+    const columns = rawData.some(d => d.expiration)
```
arielshaqed: Please always keep the same columns -- it's less surprising.

test/spec.yml
```diff
@@ -101,7 +101,7 @@
         [201*-*-*T*:*:*Z] Function invocation took * us
     - in: bn list --json
       out: |-
-        *[*{"name":"rand*","lastDeployed":"[20*-*-*T*:*:*Z]"}*]*
+        *[*{"name":"rand*","lastDeployed":"20*-*-*T*:*:*Z"}*]*
```
arielshaqed: Use `%` instead of `*`, especially for long outputs like `bn list`.  I added this to avoid #393 and similar.

test/spec.yml
```diff
@@ -227,7 +293,7 @@
              listtest2                *201*-*-*T*:*:*Z*
     -   in: bn list --json
         out: |-
-            *[*{"name":"listtest1","lastDeployed":"[20*-*-*T*:*:*Z]"}*,*{"name":"listtest2","lastDeployed":"[20*-*-*T*:*:*Z]"}*]*
+            *[*{"name":"listtest1","lastDeployed":"20*-*-*T*:*:*Z"}*,*{"name":"listtest2","lastDeployed":"20*-*-*T*:*:*Z"}*]*
```
arielshaqed: Again, these are the worst offenders in cases like #393.  Please use `#` and `%`.

michaeladda:
FYI, just marked this DO NOT MERGE, since it first needs the backend support deployed.

lib/index.js
arikmaor: done

lib/index.js
arikmaor: done

test/spec.yml
arikmaor: done in d78d791

test/spec.yml
arikmaor: done in d78d791

PR #454, opened 2019-03-14
Add experimental warning for java8

rylandg:
After a short discussion we decided to add a small warning to `bn create` for `java8` mostly driven by the fact that there are no `java8` scoobie tests.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,7 @@
 
 const { inspect } = require('util');
 const deploy = require('./deploy');
-const create = require('./create');
+const { create, experimentalRuntimes } = require('./create');
 const feedback = require('./feedback');
 const invoke = require('./invoke');
 const list = require('./list');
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -109,6 +109,9 @@ const createHandler = async function createHandler(options) {
     logger.verbose(`with ${inspect(config)}`);
   }
   logger.info(`  (use "bn deploy${optPath}${optName}" to deploy the function)`);
+  if (experimentalRuntimes.includes(options.runtime)) {
+    logger.warn(`Please be aware, support for ${options.runtime} is currently experimental`);
```
michaeladda: I would go for a shorter version, such as ` . [notice: java8 is experimental]` (also notice the two spaces).
That said, better have @avnerbraverman or @urirb take a look at the message.

rylandg:
@michaeladda will increase patch, will push them for feedback.

avnerbraverman:
I'd use something shorter:

`${options.runtime} support is experimental`

rylandg:
@avnerbraverman please take a look now

rylandg:
Actually since I opted to use your suggestion verbatim, i'll just merge this. Thanks!

urirb:
LGTM

PR #455, opened 2019-03-20
Decode body in py3 data test

bergundy:


bergundy:
CI failed due to unrelated error

PR #456, opened 2019-03-21
Add triggers cli

arikmaor:
Still needs to add options to specify different trigger source parts instead of uri

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@ define cli_envs
 	-e BINARIS_ACCOUNT_ID      \
 	-e BINARIS_INVOKE_ENDPOINT \
 	-e BINARIS_DEPLOY_ENDPOINT \
-	-e BINARIS_LOG_ENDPOINT
+	-e BINARIS_LOG_ENDPOINT    \
+	-e REDIS_PASS
 endef
 
 BRANCH := $(shell if [[ ! -z $${BRANCH_NAME+x} ]]; then echo $${BRANCH_NAME}; else git rev-parse --abbrev-ref HEAD 2>/dev/null || echo UNKNOWN; fi)
--- a/cli.js
+++ b/cli.js
@@ -8,7 +8,8 @@ const { parseTimeString } = require('./lib/timeUtil');
 const {
   deployHandler, createHandler, feedbackHandler, invokeHandler,
   listHandler, logsHandler, loginHandler, removeHandler, perfHandler,
-  showHandler, statsHandler,
+  showHandler, statsHandler, triggerListHandler, triggerUpsertHandler,
+  triggerRemoveHandler,
 } = require('./lib');
 const { forceRealm } = require('./sdk');
 
@@ -170,6 +171,66 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   }, async (argv) => {
     await handleCommand(argv, listHandler);
   })
+  .command('trigger', 'Manages triggers', yargs0 => yargs0
+      .demandCommand()
+      .command('setup', 'Creates a new trigger or updates an existing one', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 trigger setup <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {
+            demandOption: true,
+            describe: 'Redis stream URI',
+            type: 'string',
+          })
+          .option('target', {
+            demandOption: true,
+            describe: 'Function to run for each item in the stream',
+            type: 'string',
+          })
+          .option('error-handler', {
+            describe: 'Function to run when an error occurs',
+            type: 'string',
+          })
+          .demandCommand(1)
+          .strict()
+          .example(
+`  // Creates trigger 'my-trigger' that will invoke 'my-func' for every entry in 'my-stream'
+  bn trigger setup my-trigger --source-uri "redis://:redispass@redis.foo.com#my-stream" --target my-func
+
+  // Redis stream URI structure:
+  PROTOCOL   ://[:PASSWORD@]HOST       [:PORT][/DB]#STREAM_NAME
+  redis[+tls]://[:secret  @]example.com[:6379][/0] #numbers`,
+          );
+      }, async (argv) => {
+        await handleCommand(argv, triggerUpsertHandler);
+      })
+      .command('remove', 'Remove an existing trigger', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 trigger remove <trigger> [options]')
+          .positional('trigger', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .demandCommand(1)
+          .strict();
+      }, async (argv) => {
+        await handleCommand(argv, triggerRemoveHandler);
+      })
+      .command('list', 'Lists all existing triggers', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 trigger list [options]')
+          .option('json', {
+            describe: 'Output as JSON',
+            type: 'boolean',
+          })
+          .strict();
+      }, async (argv) => {
+        await handleCommand(argv, triggerListHandler);
+      })
+      .strict())
   .command('perf <function> [options]', 'Measure invocation latency (experimental)', (yargs0) => {
     yargs0
       .usage('Usage: $0 perf <function> [options]')
@@ -355,7 +416,7 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
     await handleCommand(argv, feedbackHandler);
   })
   // .strict()
-  .demand(1, 'Please provide at least 1 valid command')
+  .demandCommand(1, 'Please provide at least 1 valid command')
   .help('help')
   .epilog(`Tip:
   You can export BINARIS_LOG_LEVEL=[silly|debug|verbose] to view debug logs`)
--- a/cli.js
+++ b/cli.js
@@ -30,11 +30,6 @@ const handleCommand = async function handleCommand(options, specificHandler) {
   if (realm) {
     forceRealm(realm);
   }
-  const cmdSeq = options._;
-  // `_` is the array holding all commands given to yargs
-  if (cmdSeq.length > 1) {
-    msgAndExit(`Invalid subcommand ${cmdSeq[1]} for command ${cmdSeq[0]}`, true);
-  }
 
   await specificHandler(options);
   process.exit(0);
--- a/cli.js
+++ b/cli.js
@@ -166,7 +166,10 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   .command('list [options]', 'List all deployed functions', (yargs0) => {
     yargs0
       .usage('Usage: $0 list [options]')
-      .option('json', { describe: 'Output as JSON' })
+      .option('json', {
+        describe: 'Output as JSON',
+        type: 'boolean',
+      })
       .strict();
   }, async (argv) => {
     await handleCommand(argv, listHandler);
@@ -339,7 +342,10 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
         describe: 'Output statistics until given ISO timestamp (non-inclusive)',
         type: 'string',
       })
-      .option('json', { describe: 'Output as JSON' })
+      .option('json', {
+        describe: 'Output as JSON',
+        type: 'boolean',
+      })
       .strict()
       .example(
 `  // Retrieve all usage statistics of the account
--- a/lib/index.js
+++ b/lib/index.js
@@ -11,6 +11,7 @@ const logs = require('./logs');
 const logger = require('./logger');
 const perf = require('./perf');
 const remove = require('./remove');
+const triggers = require('./triggers');
 const YMLUtil = require('./binarisYML');
 const {
   getAllConf,
@@ -32,6 +33,8 @@ const path = require('path');
 const inquirer = require('inquirer');
 const PassThroughStream = require('stream').PassThrough;
 const columnify = require('columnify');
+
+const DEFAULT_REDIS_PORT = 6379;
 
 /**
  * Wrapper which centralizes the error handling/processing
@@ -353,6 +356,89 @@ const feedbackHandler = async function feedbackHandler(options) {
   logger.info('Thank you!');
 };
 
+async function triggerUpsertHandler(options) {
+  // options = {"_":["triggers","remove","triggerName"],"$0":"bn",...otherOptions}
+  // We have to use the position number here because of lack of good support
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
+  const { sourceUri, target, errorHandler } = options;
+  const triggerOptions = {
+    source: {
+      uri: sourceUri,
+    },
+    target: {
+      funcName: target,
+    },
+  };
+  if (errorHandler) {
+    triggerOptions.errorHandler = {
+      funcName: errorHandler,
+    };
+  }
+  const { status } = await triggers.upsert(triggerName, triggerOptions);
+  logger.info(status);
+}
+
+async function triggerRemoveHandler(options) {
+  // options = {"_":["triggers","remove","triggerName"],"$0":"bn"}
+  // We have to use the position number here because of lack of good support
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
+  const { status } = await triggers.remove(triggerName);
+  logger.info(status);
+}
+
+function formatTriggerSource(source) {
+  if (!source) {
+    return undefined;
+  }
+
+  const { proto, tls, host, port, db, streamName, password } = source;
+  const dbPart = db ? `/${db}` : '';
+  const portPart = port !== DEFAULT_REDIS_PORT ? `:${port}` : '';
+  const passwordPart = password ? `:${password}@` : '';
+  const tlsPart = tls ? '+tls' : '';
+  return `${proto}${tlsPart}://${passwordPart}${host}${portPart}${dbPart}#${streamName}`;
+}
+
+async function triggerListHandler(options) {
+  const result = await triggers.list();
+  const rawData = result.list.map(({ name, config, updatedAt }) => ({
+    name,
+    target: config.target.funcName,
+    source: formatTriggerSource(config.source),
+    errorHandler: config.errorHandler ? config.errorHandler.funcName : undefined,
+    lastUpdate: updatedAt,
+  }));
+  if (options.json) {
+    logger.info(JSON.stringify(rawData));
+  } else if (rawData.length === 0) {
+    logger.info('No triggers defined in the system.');
+  } else {
+    const minWidth = 20;
+    logger.info(columnify(rawData, {
+      columns: ['name', 'source', 'target', 'lastUpdate'],
+      config: {
+        name: {
+          headingTransform: () => 'TRIGGER',
+          minWidth,
+        },
+        source: {
+          minWidth,
+        },
+        target: {
+          minWidth,
+        },
+        lastUpdate: {
+          headingTransform: () => 'LAST UPDATE',
+        },
+      },
+    }));
+  }
+}
+
 module.exports = {
   deployHandler: exceptionWrapper(deployHandler),
   createHandler: exceptionWrapper(createHandler),
@@ -365,5 +451,7 @@ module.exports = {
   removeHandler: exceptionWrapper(removeHandler),
   statsHandler: exceptionWrapper(statsHandler),
   showHandler: exceptionWrapper(showHandler),
-  unknownHandler: exceptionWrapper(() => {}),
+  triggerListHandler: exceptionWrapper(triggerListHandler),
+  triggerUpsertHandler: exceptionWrapper(triggerUpsertHandler),
+  triggerRemoveHandler: exceptionWrapper(triggerRemoveHandler),
 };
--- a/lib/triggers.js
+++ b/lib/triggers.js
@@ -1,0 +1,25 @@
+'use strict';
+
+const { getAPIKey } = require('./userConf');
+const { triggers } = require('../sdk');
+
+async function upsert(triggerName, triggerOptions) {
+  const apiKey = await getAPIKey();
+  return triggers.upsert(apiKey, triggerName, triggerOptions);
+}
+
+async function remove(triggerName) {
+  const apiKey = await getAPIKey();
+  return triggers.remove(apiKey, triggerName);
+}
+
+async function list() {
+  const apiKey = await getAPIKey();
+  return triggers.list(apiKey);
+}
+
+module.exports = {
+  upsert,
+  remove,
+  list,
+};
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "6.6.4",
+  "version": "6.7.0",
   "description": "Binaris SDK & CLI",
   "main": "index.js",
   "bin": {
--- a/sdk/triggers.js
+++ b/sdk/triggers.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const { getTriggersListUrl, getSpecificTriggerUrl } = require('./url');
+const { getValidatedBody } = require('./handleError');
+
+function callTriggersApi(apiKey, url, method, body) {
+  const options = {
+    url,
+    headers: {
+      'X-Binaris-Api-Key': apiKey,
+    },
+    json: true,
+    body,
+  };
+  return getValidatedBody(options, method);
+}
+
+async function upsert(apiKey, triggerName, triggerOptions) {
+  return callTriggersApi(
+    apiKey,
+    getSpecificTriggerUrl(triggerName),
+    'post',
+    triggerOptions,
+  );
+}
+
+async function remove(apiKey, triggerName) {
+  return callTriggersApi(
+    apiKey,
+    getSpecificTriggerUrl(triggerName),
+    'delete',
+  );
+}
+
+async function list(apiKey) {
+  return callTriggersApi(
+    apiKey,
+    getTriggersListUrl(),
+    'get',
+  );
+}
+
+module.exports = {
+  upsert,
+  remove,
+  list,
+};
--- a/sdk/url.js
+++ b/sdk/url.js
@@ -31,6 +31,14 @@ const getStatsUrl = function getStatsUrl(accountId) {
   return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'metrics', accountId);
 };
 
+function getTriggersListUrl() {
+  return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'triggers');
+}
+
+function getSpecificTriggerUrl(triggerName) {
+  return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'triggers', triggerName);
+}
+
 module.exports = {
   getCodeUploadUrl,
   getConfUploadUrl,
@@ -39,4 +47,6 @@ module.exports = {
   getLogsUrl,
   getListUrl,
   getStatsUrl,
+  getTriggersListUrl,
+  getSpecificTriggerUrl,
 };
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -13,6 +13,7 @@
                 bn remove <function> [options]            Remove a previously deployed function
                 bn invoke <function> [options]            Invoke a Binaris function
                 bn list [options]                         List all deployed functions
+                bn trigger                                Manages triggers
                 bn perf <function> [options]              Measure invocation latency (experimental)
                 bn logs <function> [options]              Print the logs of a function
                 bn show [config]                          Show Binaris account configuration
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -1334,3 +1334,213 @@
     -  in: bn create node8 testArrayOpts --config=aaa --config=bbb
        exit: 1
        err: "Non object create configuration options: aaa,bbb"
+
+- test: Test new trigger (good-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+      out: CREATED
+
+- test: Test update trigger (good-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+    - bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream2" --target $FUNC_NAME
+      out: UPDATED
+
+- test: Test 2 new triggers, same source, different target (good-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export FUNC_NAME2=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME2=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+    - |-
+        cat <<EOF >/home/dockeruser/test/binaris.yml
+        functions:
+          $FUNC_NAME:
+            file: function.js
+            entrypoint: handler
+            executionModel: concurrent
+            runtime: node8
+          $FUNC_NAME2:
+            file: function.js
+            entrypoint: handler
+            executionModel: concurrent
+            runtime: node8
+        EOF
+    - bn deploy $FUNC_NAME2
+    - bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn trigger remove $TRIGGER_NAME2
+    - bn remove $FUNC_NAME
+    - bn remove $FUNC_NAME2
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME2 --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME2
+      out: CREATED
+
+- test: Test new trigger with bad source uri (bad-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+  cleanup:
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME --source-uri redis://not-a-valid-redis-uri --target $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Could not connect to source*
+
+- test: Test new trigger with missing target function (bad-path)
+  steps:
+    - in: bn trigger setup notexistingfunctrigger --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target nonexistingfunc
+      exit: 1
+      err: |-
+        RequestId: *
+        Error: No such function
+
+- test: Test new trigger when it already exists with different name (bad-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+    - bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup testduplicatetriggerthatwillnotbecreated --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Bad Request, errors: Trigger from source to target already exists
+
+- test: Test update trigger when it already exists with different name (bad-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME=rand$RANDOM$RANDOM
+    - export TRIGGER_NAME2=rand$RANDOM$RANDOM
+    - bn create node8 $FUNC_NAME
+    - bn deploy $FUNC_NAME
+    - bn trigger setup $TRIGGER_NAME --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+    - bn trigger setup $TRIGGER_NAME2 --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream2" --target $FUNC_NAME
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn trigger remove $TRIGGER_NAME2
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME2 --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Bad Request, errors: Trigger from source to target already exists
+
+- test: Test remove trigger (good-path)
+  setup:
+    - export FUNC_NAME=rand$RANDOM$RANDOM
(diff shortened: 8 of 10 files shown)


lib/index.js
```diff
@@ -353,6 +354,71 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+}
+
+async function triggerRemoveHandler(options) {
+  // options = {"_":["triggers","remove","triggerName"],"$0":"bn"}
+  // We have to use the position number here because of lack of good support
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
+  const result = await triggers.remove(triggerName);
+  logger.info(options.json ? JSON.stringify(result) : result.status);
```
bergundy: I don't think we need json output here

lib/index.js
```diff
@@ -353,6 +354,71 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+    logger.info(JSON.stringify(rawData));
+  } else if (rawData.length > 0) {
+    logger.info(columnify(rawData, {
+      columns: ['name', 'target', 'errorHandler', 'lastUpdate'],
+      config: {
+        name: {
+          headingTransform: () => 'TRIGGER',
+          minWidth: 25,
+        },
+        target: {
```
bergundy: What about showing `source`? password is already redacted by server

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   }, async (argv) => {
     await handleCommand(argv, listHandler);
   })
+  .command('triggers', 'Manages triggers', yargs0 => yargs0
+      .demandCommand()
+      .command('upsert', 'Creates a new trigger or updates an existing one', (yargs1) => {
```
michaeladda: not sure about the name upsert.
i think it's better than either create up update, yet it may be confusing.

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        yargs1
+          .usage('Usage: $0 triggers upsert <trigger name> [options]')
+          .positional('trigger name', {
```
michaeladda: do we need "name" here?
in comparison functions print `bn deploy <function>` not `bn deploy <function name>`

sdk/url.js
```diff
@@ -59,6 +59,14 @@ const getStatsUrl = function getStatsUrl(accountId, apiKey) {
   return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'metrics', accountId);
 };
 
+function getTriggersListUrl() {
+  return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'triggers');
+}
+
+function getSpecificTriggersUrl(triggerName) {
```
bergundy: nit: `Trigger` in singular

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   }, async (argv) => {
     await handleCommand(argv, listHandler);
   })
+  .command('triggers', 'Manages triggers', yargs0 => yargs0
```
michaeladda: I think an example section is needed here to show a very brief flow

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {
+            demandOption: true,
+            describe: 'Redis-stream uri',
+            type: 'string',
+          })
+          .option('target-func', {
```
michaeladda: why not simply "function"?

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+  .command('triggers', 'Manages triggers', yargs0 => yargs0
+      .demandCommand()
+      .command('upsert', 'Creates a new trigger or updates an existing one', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers upsert <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {
```
michaeladda: maybe just "source"?

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+            demandOption: true,
+            describe: 'Redis-stream uri',
+            type: 'string',
+          })
+          .option('target-func', {
+            demandOption: true,
+            describe: 'Name of the function to run for each item in the stream',
+            type: 'string',
+          })
+          .option('error-handler-func', {
```
michaeladda: consider simply "error".
also, this should be a hidden option imho, at least initially

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        await handleCommand(argv, triggerUpsertHandler);
+      })
+      .command('remove', 'Removes an existing trigger', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers remove <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {})
```
michaeladda: why is the source needed?
if it is needed, type should be described

cli.js
```diff
@@ -170,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+          .option('target-func', {
+            demandOption: true,
+            describe: 'Name of the function to run for each item in the stream',
+            type: 'string',
+          })
+          .option('error-handler-func', {
+            describe: 'Name of the function to run when an error occurres',
+            type: 'string',
+          })
+          .demand(1)
```
michaeladda: `demand` is deprecated according to https://github.com/yargs/yargs/blob/master/docs/api.md#demand.
can this be replaced with `demandOption` or `demandCommand`?

lib/index.js
```diff
@@ -353,6 +354,71 @@ const feedbackHandler = async function feedbackHandler(options) {
   logger.info('Thank you!');
 };
 
+async function triggerUpsertHandler(options) {
+  // We have to use the position number here because of lack of good support
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
```
michaeladda: smelly indeed 💩 
at least the comment help

lib/index.js
```diff
@@ -353,6 +354,71 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+      funcName: options.targetFunc,
+    },
+  };
+  if (options.errorHandlerFunc) {
+    triggerOptions.errorHandler = {
+      funcName: options.errorHandlerFunc,
+    };
+  }
+  const result = await triggers.upsert(triggerName, triggerOptions);
+  logger.info(options.json ? JSON.stringify(result) : result.status);
```
michaeladda: why stringifying the entire result but printing only status?

lib/index.js
```diff
@@ -353,6 +354,71 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
+  const result = await triggers.remove(triggerName);
+  logger.info(options.json ? JSON.stringify(result) : result.status);
+}
+
+async function triggerListHandler(options) {
+  const result = await triggers.list();
+  const rawData = (result.list || []).map(({ name, config, updatedAt }) => ({
```
michaeladda: i wish this was typescript too :-(

cli.js
```diff
@@ -175,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+        await handleCommand(argv, triggerUpsertHandler);
+      })
+      .command('remove', 'Removes an existing trigger', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers remove <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {})
```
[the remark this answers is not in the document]
arikmaor: Removing

cli.js
```diff
@@ -175,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+          .demand(1)
+          .strict();
+      }, async (argv) => {
+        await handleCommand(argv, triggerRemoveHandler);
+      })
+      .command('list', 'Lists all existing triggers', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers list [options]')
+          .option('json', { describe: 'Output as JSON' })
```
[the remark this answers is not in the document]
arikmaor: It's in a couple of other places (I've copied this line)

changing them all

lib/index.js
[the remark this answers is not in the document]
arikmaor: Mistake, Changing to:
```js
const { status } = await triggers.upsert(triggerName, triggerOptions); logger.info(status);
```

lib/index.js
arikmaor: Mistake, Changing to:
```js
const { status } = await triggers.remove(triggerName);
logger.info(status);
```

lib/index.js
arikmaor: sure lets add it, but lets decide exactly what will be displayed
only `host` and `streamName`?
```
TRIGGER              SOURCE HOST     		SOURCE STREAM   TARGET               ERROR HANDLER        LAST UPDATE
test2                misc-redis.binaris.io  number		    hello                                     2019-03-21T09:38:55.875Z
```

sdk/url.js
arikmaor: Fixing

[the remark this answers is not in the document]
arikmaor: Fixing

lib/index.js
```diff
@@ -365,5 +431,7 @@ module.exports = {
   removeHandler: exceptionWrapper(removeHandler),
   statsHandler: exceptionWrapper(statsHandler),
   showHandler: exceptionWrapper(showHandler),
-  unknownHandler: exceptionWrapper(() => {}),
```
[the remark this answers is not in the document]
arikmaor: It's unused

cli.js
arikmaor: lets try to think about a better name, maybe @rylandg or @avnerbraverman have a suggestion.
Leaving it as is for now

cli.js
arikmaor: Changing

cli.js
arikmaor: Removing

cli.js
bergundy: It would be nice to support the verbose trigger format, uri is just one way to get the source in

lib/index.js
bergundy: When would list not exist?

lib/index.js
bergundy: You'd need proto too, we'd like to support other triggers except for redis.
I'd have one column for `source` and create a formatter for each source.
For redis://:secret@host:666/0#streamName` you could print: `redis://host:666/0#streamName`
If port is the default redis port (6379 you could omit it).

cli.js
arikmaor: Why it should be hidden, @bergundy are we not gonna implement it before redis conf?

lib/index.js
arikmaor: Removing `json` option

lib/index.js
arikmaor: it's not supposed to, still pretty risky leaving that out...

bergundy: It shouldn't be, we have type validation on all of our APIs, if we were using the generated interfaces, which we can't just yet, you would have seen that this is unnecessary.

lib/index.js
arikmaor: so maybe this should also come from the server? (as he is the one who does the parsing, he should know how to build the uri)

arikmaor: Fixed so not output looks like that
```
TRIGGER         SOURCE                                           TARGET               LAST UPDATE
test2           redis+tls://misc-redis.binaris.io:6379/0#numbers hello                2019-03-21T09:38:55.875Z
```

Also removed the error handler as it's not implemented yet, left it as is when using `--json` flag

lib/index.js
arikmaor: so... are you saying I should remove it?

cli.js
arikmaor: Changing all `demand` occurrences to use `demandCommand` instead

cli.js
```diff
@@ -175,6 +171,56 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+      .command('upsert', 'Creates a new trigger or updates an existing one', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers upsert <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {
+            demandOption: true,
+            describe: 'Redis-stream uri',
```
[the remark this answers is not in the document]
arikmaor: Add examples, but not sure if it's good enough :\

cli.js
arikmaor: Added example for `upsert` command, `list` and `remove` are trivial
Not sure my example is good enough though...

cli.js
arikmaor: lets try to get @avnerbraverman opinion (as this is a product issue) before we make 10k iteration over flag names...

sdk/url.js
```diff
@@ -59,6 +59,14 @@ const getStatsUrl = function getStatsUrl(accountId, apiKey) {
   return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'metrics', accountId);
 };
 
+function getTriggersListUrl() {
+  return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'triggers');
+}
+
+function getSpecificTriggerUrl(triggerName) {
+  return urljoin(`https://${getDeployEndpoint()}`, 'v3', 'trigger', triggerName);
```
rylandg: It might be slightly drier to have a base endpoint function for just the `https://${getDeployEndpoint()}`, 'v3' part

lib/triggers.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const { getAPIKey } = require('./userConf');
+const { triggers } = require('../sdk');
+
+async function upsert(triggerName, triggerOptions) {
```
rylandg: nit: this code resides inside `triggers.js` no need to prefix every variable with "trigger"

sdk/triggers.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const { getTriggersListUrl, getSpecificTriggerUrl } = require('./url');
+const { getValidatedBody } = require('./handleError');
+
+async function upsert(apiKey, triggerName, triggerOptions) {
+  const listOptions = {
+    url: getSpecificTriggerUrl(triggerName),
```
rylandg: I would name it `getTriggerSpecificUrl` but it's not a big deal

sdk/triggers.js
```diff
@@ -0,0 +1,44 @@
...
+    headers: {
```
rylandg: This can be defined once for all trigger actions

sdk/triggers.js
```diff
@@ -0,0 +1,44 @@
...
+      'X-Binaris-Api-Key': apiKey,
+    },
+    json: true,
+    body: triggerOptions,
+  };
+  return getValidatedBody(listOptions);
+}
+
+async function remove(apiKey, triggerName) {
+  const listOptions = {
```
rylandg: They are all called `listOptions` even for `remove` and `upsert`.

sdk/triggers.js
```diff
@@ -0,0 +1,44 @@
...
+    url: getTriggersListUrl(),
+    headers: {
+      'X-Binaris-Api-Key': apiKey,
+    },
+    json: true,
+  };
+  return getValidatedBody(listOptions, 'get');
+}
+
+module.exports = {
```
rylandg: This entire file could a single function that takes in like 3 params and generates a new function.

cli.js
bergundy: It's supposed to be `bn trigger`, singular: https://github.com/binaris/specs/blob/master/Redis%20Streams%20Trigger.md#trigger-configuration

arikmaor: changed to `trigger`

sdk/triggers.js
arikmaor: changed all of them to just `options`

sdk/triggers.js
arikmaor: 👍

sdk/triggers.js
arikmaor: refactored the file a bit, you can take a look

cli.js
```diff
@@ -30,11 +31,6 @@ const handleCommand = async function handleCommand(options, specificHandler) {
   if (realm) {
     forceRealm(realm);
   }
-  const cmdSeq = options._;
```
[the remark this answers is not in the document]
arikmaor: This code gives an error in case we have an extra command, for example:
`bn list something-else`
without this code, `yargs` will just ignore it

this code relies on the fact that we don't have subcommands, which is not true now

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
   .command('list [options]', 'List all deployed functions', (yargs0) => {
     yargs0
       .usage('Usage: $0 list [options]')
-      .option('json', { describe: 'Output as JSON' })
+      .option('json', {
+        describe: 'Output as JSON',
+        type: 'boolean',
```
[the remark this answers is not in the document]
arikmaor: there is not default:
```
type: 'boolean' =>
  --json      Output as JSON  [boolean]

type: 'string' =>
  --json      Output as JSON  [string]

nothing =>
  --json      Output as JSON
```

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+`  // Setup a trigger called 'my-trigger' that will invoke 'my-func' for every entry in 'my-stream'
+  bn triggers upsert my-trigger --source-uri "redis://:redis-password@my-redis.my-domain.com/0#my-stream" --target-func my-func
+
+  // Same this but secured
+  bn triggers upsert my-trigger --source-uri "redis+tls://:redis-password@my-redis.my-domain.com/0#my-stream" --target-func my-func`,
+          );
+      }, async (argv) => {
+        await handleCommand(argv, triggerUpsertHandler);
+      })
+      .command('remove', 'Removes an existing trigger', (yargs1) => {
```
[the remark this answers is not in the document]
arikmaor: changed

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+          .positional('trigger', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .demandCommand(1)
+          .strict();
+      }, async (argv) => {
+        await handleCommand(argv, triggerRemoveHandler);
+      })
+      .command('list', 'Lists all existing triggers', (yargs1) => {
```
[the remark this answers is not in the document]
arikmaor: are you suggesting there will be this commands:
```
bn list // lists functions
bn list --triggers // lists triggers
```
?

(I don't like it...)

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
   logger.info('Thank you!');
 };
 
+async function triggerUpsertHandler(options) {
+  // We have to use the position number here because of lack of good support
```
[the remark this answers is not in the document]
arikmaor: When doing a subcommand, I can't get the positional arguments in the nice way I can get it on the root command,
thats why I need to extract them manually from `options._`

I haven't tried `nested-yargs`, but it's one file

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
+  const triggerOptions = {
```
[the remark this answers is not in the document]
arikmaor: changing

lib/index.js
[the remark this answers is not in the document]
arikmaor: Do you see an issue there?

cli.js
arikmaor: Ack

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+            describe: 'Redis-stream uri',
+            type: 'string',
+          })
+          .option('target-func', {
+            demandOption: true,
+            describe: 'Name of the function to run for each item in the stream',
+            type: 'string',
+          })
+          .option('error-handler-func', {
+            describe: 'Name of the function to run when an error occurres',
```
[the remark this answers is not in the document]
arikmaor: Changed

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+  .command('triggers', 'Manages triggers', yargs0 => yargs0
+      .demandCommand()
+      .command('upsert', 'Creates a new trigger or updates an existing one', (yargs1) => {
+        yargs1
+          .usage('Usage: $0 triggers upsert <trigger name> [options]')
+          .positional('trigger name', {
+            describe: 'Trigger name',
+            type: 'string',
+          })
+          .option('source-uri', {
```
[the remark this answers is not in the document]
arikmaor: I think after merging this PR, I'll do another one just for the interfaces

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+            type: 'string',
+          })
+          .option('error-handler-func', {
+            describe: 'Name of the function to run when an error occurres',
+            type: 'string',
+          })
+          .demandCommand(1)
+          .strict()
+          .example(
+`  // Setup a trigger called 'my-trigger' that will invoke 'my-func' for every entry in 'my-stream'
```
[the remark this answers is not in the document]
arikmaor: Changed

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+  bn triggers upsert my-trigger --source-uri "redis://:redis-password@my-redis.my-domain.com/0#my-stream" --target-func my-func
```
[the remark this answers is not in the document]
arikmaor: Changing, but there is little difference between to two...
```
redis://:redis-password@my-redis.my-domain.com/0#my-stream
redis://:redispass@redis.foo.com/0#my-stream
```

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+    triggerOptions.errorHandler = {
+      funcName: options.errorHandlerFunc,
+    };
+  }
+  const { status } = await triggers.upsert(triggerName, triggerOptions);
+  logger.info(status);
+}
+
+async function triggerRemoveHandler(options) {
+  // options = {"_":["triggers","remove","triggerName"],"$0":"bn"}
```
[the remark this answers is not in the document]
arikmaor: refactored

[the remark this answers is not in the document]
arikmaor: 1. There are 2 places, I hope we will remove them as we are considering writing the whole cli all together
2. Did understand the comment

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+  // We have to use the position number here because of lack of good support
+  // for nested commands with positional arguments
+  // consider replacing with `nested-yargs` npm package
+  const triggerName = options._[2];
```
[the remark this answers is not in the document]
arikmaor: Ack

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+    return undefined;
+  }
+
+  const { proto, tls, host, port, db, streamName } = source;
+  return `${proto}${tls ? '+tls' : ''}://${host}:${port}/${db}#${streamName}`;
+}
+
+async function triggerListHandler(options) {
+  const result = await triggers.list();
+  const rawData = (result.list || []).map(({ name, config, updatedAt }) => ({
```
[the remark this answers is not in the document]
arikmaor: why is this better?
you are suggesting to change
```js
list.map(({ name, config, updatedAt }) => ({
  lastUpdate: updatedAt
})
```
with
```js
list.map(({ name, config, updatedAt: lastUpdate }) => ({
  lastUpdate
})
```
IMHO the first is more readable, but that's just a personal opinion...

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+    name,
+    target: config.target.funcName,
+    source: formatTriggerSource(config.source),
+    errorHandler: config.errorHandler ? config.errorHandler.funcName : undefined,
+    lastUpdate: updatedAt,
+  }));
+  if (options.json) {
+    logger.info(JSON.stringify(rawData));
+  } else if (rawData.length > 0) {
```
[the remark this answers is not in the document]
arikmaor: Didn't understand the question

lib/index.js
```diff
@@ -353,6 +354,80 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+    logger.info(columnify(rawData, {
+      columns: ['name', 'source', 'target', 'lastUpdate'],
+      config: {
+        name: {
+          headingTransform: () => 'TRIGGER',
+          minWidth: 20,
```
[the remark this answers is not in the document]
arikmaor: changed

sdk/url.js
arikmaor: don't want to refactor the whole cli right now :\

lib/triggers.js
arikmaor: I think it's good in this case, because it's still very short, and it makes it easy to read then copying to conversations

arikmaor: also, `name` alone could mean lots of things

`options` could suggest (to me at least) that this options could include things like the `--json` flag, this would not be an issue if you use typescript

sdk/triggers.js
arikmaor: Ack

cli.js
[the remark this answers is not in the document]
rylandg: I understand now, thanks for explaining your rationale. That being said, I wrote the code for the problem you described. It seems like a bad user experience to not error on superfluous extra commands. Can you either

a) keep the existing functionality and add logic for special exceptions (as with triggers)
b) at minimum open an issue for a)

cli.js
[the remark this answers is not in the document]
rylandg: I understand, I believe the implicit default is "boolean". I'll never complain about more clarity though.

cli.js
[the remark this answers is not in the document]
rylandg: If that's the plan, make an issue now please.

cli.js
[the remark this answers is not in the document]
rylandg: I'm just saying that right now you don't type 

`bn functions --list`  you type `bn list`

But for triggers you type `bn triggers --list`

Maybe we decide that because functions are a first class citizen they should have a different UX. But as of now, the interface might seem inconsistent to a new user.

FTR: We could always make the default of `bn list` "functions" so `bn list` would still print functions but `bn list --triggers` would print triggers.

lib/index.js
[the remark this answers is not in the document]
rylandg: Will look into it.

lib/index.js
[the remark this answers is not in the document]
rylandg: Cool, thanks.

lib/index.js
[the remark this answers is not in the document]
rylandg: 1.  Don't understand but ok.
2. The functions have identical signatures but different comments.

lib/index.js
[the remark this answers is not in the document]
rylandg: When someone doesn't request JSON you only print if the `rawData` array has items. But if they request JSON you always print the array even with 0 items.

lib/triggers.js
rylandg: Good points.

lib/index.js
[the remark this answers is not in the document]
rylandg: Yea, I should have caught this before. Nice catch.

sdk/triggers.js
rylandg: much better, thanks

sdk/url.js
rylandg: I don't know how that requires a whole CLI refactor, but it's minimal anyway

cli.js
bergundy: Setup / apply / deploy are good options

lib/index.js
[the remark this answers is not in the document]
arikmaor: :)

sdk/url.js
arikmaor: sorry, I've miss read your comment

cli.js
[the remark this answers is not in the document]
arikmaor: Taking another look, `yargs` ignores `bn list something-else` with or without this code, without an error message
So I think it is broken anyway...

cli.js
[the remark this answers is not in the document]
arikmaor: there is no `bn triggers --list`
it's `bn trigger list`

Makefile
```diff
@@ -18,7 +18,8 @@ define cli_envs
 	-e BINARIS_ACCOUNT_ID      \
 	-e BINARIS_INVOKE_ENDPOINT \
 	-e BINARIS_DEPLOY_ENDPOINT \
-	-e BINARIS_LOG_ENDPOINT
+	-e BINARIS_LOG_ENDPOINT    \
+	-e REDIS_PASS
```
bergundy: nit: `REDIS_PASSWORD`

cli.js
```diff
@@ -30,11 +31,6 @@ const handleCommand = async function handleCommand(options, specificHandler) {
   if (realm) {
     forceRealm(realm);
```
bergundy: Sidenote, do we still need this?

cli.js
bergundy: @arikmaor please rename to one of the options above, I like `setup`.

cli.js
bergundy: Not changed

cli.js
bergundy: There's also an error handler, I vote for calling it `target` and dropping the `func` part.

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+            demandOption: true,
+            describe: 'Redis-stream uri',
+            type: 'string',
+          })
+          .option('target-func', {
+            demandOption: true,
+            describe: 'Function to run for each item in the stream',
+            type: 'string',
+          })
+          .option('error', {
```
bergundy: Not a good name, should be `error-handler`.

lib/index.js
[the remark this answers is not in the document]
bergundy: It would be better to print `No triggers` than not printing anything.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "6.6.4",
+  "version": "6.7.0-rc1",
```
bergundy: Please change before merging.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
     -  in: bn create node8 testArrayOpts --config=aaa --config=bbb
        exit: 1
        err: "Non object create configuration options: aaa,bbb"
+
+- test: Test new trigger (good-path)
+  cleanup:
+    - bn trigger remove testtrigger
+    - bn remove testtriggerfunc
+  steps:
+    - in: bn create node8 testtriggerfunc
```
bergundy: Better create all functions with ttl, it's easier than garbage collection.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+      out: |-
```
bergundy: Not worth testing, it's tested in the functions tests.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+          Created function * in /home/dockeruser/test
+            (use "bn deploy testtriggerfunc" to deploy the function)
+    - in: bn deploy testtriggerfunc
+      out: |-
```
bergundy: Same.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+
+- test: Test update trigger (good-path)
+  cleanup:
+    - bn trigger remove testupdatetrigger
+    - bn remove testupdatetriggerfunc
+    - bn remove testupdatetriggerfunc2
+  steps:
+    - in: bn create node8 testupdatetriggerfunc
+      out: |-
+          Created function * in /home/dockeruser/test
```
bergundy: Don't test this here please.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+        EOF
+    - in: bn deploy testupdatetriggerfunc2
+      out: |-
+        Deployed function testupdatetriggerfunc2
+        Invoke with one of:
+          "bn invoke testupdatetriggerfunc2"
+          "curl -H X-Binaris-Api-Key:* https://*"
+    - in: bn trigger upsert testupdatetrigger --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#numbers" --target-func testupdatetriggerfunc
+      out: CREATED
+    - in: bn trigger upsert testupdatetrigger --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#numbers" --target-func testupdatetriggerfunc2
```
bergundy: Could have just changed the URI to use a different stream name instead of changing the function, would save the time of deploying and removing the second function.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+      out: |-
+        Deployed function testbadtriggerfunc
+        Invoke with one of:
+          "bn invoke testbadtriggerfunc"
+          "curl -H X-Binaris-Api-Key:* https://*"
+    - in: bn trigger upsert testbadtrigger --source-uri redis://not-a-valid-redis-uri --target-func testbadtriggerfunc
+      exit: 1
+      err: |-
+        RequestId: *
+        Internal Binaris server error*
```
bergundy: Not the expected error here.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+
+- test: Test update trigger with bad source uri (bad-path)
```
bergundy: Redundant test.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+
+- test: Test new trigger with missing target function (bad-path)
+  steps:
+    - in: bn trigger upsert notexistingfunctrigger --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#numbers" --target-func nonexistingfunc
+      exit: 1
+      err: |-
+        RequestId: *
+        Error: No such function
+
+- test: Test update trigger with missing target function (bad-path)
```
bergundy: Redundant test.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+        Invoke with one of:
+          "bn invoke testduplicatetriggerfunc"
+          "curl -H X-Binaris-Api-Key:* https://*"
+    - in: bn trigger upsert testduplicatetrigger --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#numbers" --target-func testduplicatetriggerfunc
+      out: CREATED
+    - in: bn trigger upsert testduplicatetriggerthatwillnotbecreated --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#numbers" --target-func testduplicatetriggerfunc
+      exit: 1
+      err: |-
+        RequestId: *
+        Bad Request
```
bergundy: Not the best error string, we should say that a trigger already exists for source / target combination.

test/spec.yml
```diff
@@ -1373,3 +1374,346 @@
...
+
+- test: Test update trigger when it already exists with different name (bad-path)
```
bergundy: Redundant.

bergundy:
One last thing, the most important test was left out, please test invocation via redis stream.

Makefile
arikmaor: I don't think I'll change it now as it involves changes to `precious` and staff...
Why is this more correct?

cli.js
arikmaor: can be useful...

cli.js
arikmaor: changed to setup

cli.js
arikmaor: leaving it `trigger name` as you agreed (me and @bergundy) that it's the best option.
`name` isn't clear enough
`trigger` alone looks weird

cli.js
arikmaor: changed to `target`

cli.js
arikmaor: changed

lib/index.js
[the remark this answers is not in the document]
arikmaor: copied the `options = ...` line to the other comment

lib/index.js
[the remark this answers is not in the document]
arikmaor: Add message `No triggers defined in the system.` @rylandg

test/spec.yml
arikmaor: yet it passes...
this needs to be updated on `spice`

test/spec.yml
arikmaor: yet it passes...
this needs to be updated on `spice`

test/spec.yml
arikmaor: removed

test/spec.yml
arikmaor: remove

test/spec.yml
arikmaor: removed

test/spec.yml
arikmaor: Why?

test/spec.yml
arikmaor: Why?

test/spec.yml
arikmaor: Why?

test/spec.yml
arikmaor: Ack

arikmaor:
I've add a redis invoke test and randomized all generated function names @bergundy

Makefile
bergundy: I just think it's clearer, not critical

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+          })
+          .option('error-handler', {
+            describe: 'Function to run when an error occurs',
+            type: 'string',
+          })
+          .demandCommand(1)
+          .strict()
+          .example(
+`  // Creates trigger 'my-trigger' that will invoke 'my-func' for every entry in 'my-stream'
+  bn trigger setup my-trigger --source-uri "redis://:redispass@redis.foo.com/0#my-stream" --target my-func
```
bergundy: `/0` is the default, you can omit it.

cli.js
```diff
@@ -170,11 +166,73 @@ Usage: $0 <command> [options]` // eslint-disable-line comma-dangle
...
+
+  // Same this but secured
+  bn trigger setup my-trigger --source-uri "redis+tls://:redispass@redis.foo.com/0#my-stream" --target my-func`,
```
bergundy: Need an explanation, like so:
```
PROTOCOL://:PASSWORD@HOST       :PORT/DB#STREAM_NAME
redis   ://:secret  @example.com:6379/0 #numbers
```

Where PORT, PASSWORD and DB are all optional.

lib/index.js
```diff
@@ -353,6 +354,85 @@ const feedbackHandler = async function feedbackHandler(options) {
...
+  logger.info(status);
+}
+
+function formatTriggerSource(source) {
+  if (!source) {
+    return undefined;
+  }
+
+  const { proto, tls, host, port, db, streamName } = source;
+  return `${proto}${tls ? '+tls' : ''}://${host}:${port}/${db}#${streamName}`;
```
bergundy: omit port if 6379, omit db if 0, show password if exists (will be XXX when returned by the server).

package.json
bergundy: Ping

test/spec.yml
bergundy: Because it's already tested on create, this test is meaningless.

test/spec.yml
bergundy: Already tested for create, it's just protocol validation which is the same for update or create.

test/spec.yml
bergundy: Check the error returned by spice, it should contain the reason, this is bad UX.

test/spec.yml
bergundy: Scratch that

test/spec.yml
```diff
@@ -1333,3 +1334,247 @@
...
+  cleanup:
+    - bn trigger remove $TRIGGER_NAME
+    - bn trigger remove $TRIGGER_NAME2
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup $TRIGGER_NAME2 --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io/0#teststream" --target $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Bad Request
```
bergundy: Should print why (should come from spice).

test/spec.yml
bergundy: You should get the correct error now

test/spec.yml
```diff
@@ -1333,3 +1334,247 @@
...
+    - bn trigger remove $TRIGGER_NAME
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn remove $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Function has associated trigger(s), remove trigger(s) first
+
+- test: Test trigger is invoked from redis stream
```
bergundy: Nice!

cli.js
arikmaor: removed it from tests as well

lib/index.js
arikmaor: done

cli.js
arikmaor: done

test/spec.yml
arikmaor: 😃

test/spec.yml
```diff
@@ -1333,3 +1334,213 @@
...
+    - bn trigger remove $TRIGGER_NAME
+    - bn remove $FUNC_NAME
+  steps:
+    - in: bn trigger setup testduplicatetriggerthatwillnotbecreated --source-uri "redis+tls://:${REDIS_PASS}@misc-redis.binaris.io#teststream" --target $FUNC_NAME
+      exit: 1
+      err: |-
+        RequestId: *
+        Bad Request, errors: Trigger from source to target already exists
+
+- test: Test update trigger when it already exists with different name (bad-path)
```
bergundy: Come to think about this again, this test is redundant, it's already tested in the test above.

PR #458, opened 2019-03-26
Use ava assert method for power-assert context

vogre:
As noticed earlier today context disappeared from assertions due to ava upgrade, this brings the context back.

Second run purposefully broken to see the errors.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -112,8 +112,8 @@ function createTest(rawSubTest) {
     // eslint-disable-next-line no-param-reassign
     t.context.cleanup = rawSubTest.cleanup;
 
-    const matchText = function matchText(expected, text) {
-      return createRegTest(expected).test(text);
+    const matchText = function matchText(expected, stripCmd, prop) {
+      return createRegTest(expected[prop]).test(stripCmd[`std${prop}`]);
     };
 
     for (const step of rawSubTest.steps) {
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -112,22 +112,22 @@ function createTest(rawSubTest) {
...
+      return createRegTest(expected[prop]).test(text[prop]);
     };
 
     for (const step of rawSubTest.steps) {
       // eslint-disable-next-line no-await-in-loop
       const cmdOut = await t.context.ct.streamIn(step.in);
       if (step.out) {
-        const stripStdout = stripText(cmdOut.stdout);
-        t.true(matchText(step.out, stripStdout));
+        const stripCmdOut = { ...cmdOut, out: stripText(cmdOut.stdout) };
```
vogre: Nit: this means that the power-assert will have both `stdout` and `out`, which are almost the same, except for whitespace.

arikmaor: Changed it, `matchText` is a little weird now but should work
Will make https://ci.binaris.io/blue/organizations/jenkins/gobinaris%2Fprecious/detail/master/7110/pipeline fail again to see the errors

test/test.js
```diff
@@ -112,22 +112,22 @@ function createTest(rawSubTest) {
...
+      return createRegTest(expected[prop]).test(stripCmd[`std${prop}`]);
     };
 
     for (const step of rawSubTest.steps) {
       // eslint-disable-next-line no-await-in-loop
       const cmdOut = await t.context.ct.streamIn(step.in);
       if (step.out) {
-        const stripStdout = stripText(cmdOut.stdout);
-        t.true(matchText(step.out, stripStdout));
+        const stripCmdOut = { ...cmdOut, stdout: stripText(cmdOut.stdout) };
```
arielshaqed: This is very nice!  It occurs to me that it might be easier to read if we gave up on `power-assert` (sorry...) and just used `t.deepEqual`, something like
```js
const cmdToCompare = {
  stdout: step.out && cmdOut.stdout,
  stderr: step.err && cmdOut.stderr,
  exit: cmdOut.exitCode,
};
const expected =
  stdout: step.out,
  stderr: step.err,
  exitCode: step.exit || 0,
};
t.deepEqual(expected, cmdToCompare);
```

arikmaor: Ack (as you say 😎)

PR #459, opened 2019-03-28
Remove api key support

bergundy:


sdk/auth.js
```diff
@@ -29,33 +29,8 @@ const verifyAPIKey = async function verifyAPIKey(apiKey) {
     };
   } catch (accountErr) {
     logger.debug('Failed to authenticate via account', { error: inspect(accountErr) });
```
vogre: `via account` can be omitted.

bergundy:
Good catch!

PR #460, opened 2019-03-28
Fix types and add trigger context attributes

bergundy:


lib/functionTemplates/js/binaris.d.ts
```diff
@@ -1,31 +1,33 @@
 /// <reference types="node" />
-interface BinarisHTTPRequest {
-  env: Record<string, string | undefined>;
-  query: Record<string, string | string[] | undefined>;
+export interface HTTPRequest {
+  env: Record<string, string>;
```
arielshaqed: This doesn't match [the type in `bolt`](https://git.io/fjkwF).

lib/functionTemplates/js/binaris.d.ts
```diff
@@ -1,31 +1,33 @@
...
+  query: Record<string, string | string[]>;
```
arielshaqed: This also doesn't match [the type in `bolt`](https://git.io/fjkwb).

lib/functionTemplates/js/binaris.d.ts
```diff
@@ -1,31 +1,33 @@
...
   body: Buffer;
   path: string;
   method: string;
   requestId: string;
-  headers: Record<string, string | string[] | undefined>;
+  headers: Record<string, string | string[]>;
```
arielshaqed: Ditto.

bergundy:
@arielshaqed PTAL.
Reverted the changes that made it incompatible with bolt.
I think the naming is much better like this.
Agree on adding tests for the types and unifying with bolt's types (#469).

PR #464, opened 2019-03-31
Use meaningful names and random components on all tests in spec

### LINKED ISSUE #462: Need better naming for function created during spec test
one idea was to inject `$PREFIX` environment variable that would hold the first 10-20 characters on the test name

arielshaqed:
Avoids reusing leftover functions, while making it easier to debug
which test is leaving them over.

Fixes #462

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -297,8 +297,8 @@
     -   in: bn list
         out: |-
              FUNCTION                 *LAST DEPLOYED*
-             listtest1                *20#-#-#T#:#:#.#Z*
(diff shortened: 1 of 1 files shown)


arielshaqed:
Thanks! 

I'm _not_ touching anything that doesn't get deployed.  OTOH, the "list" and "long names" tests should be easy enough to do in the same manner.  Got rid of much of the setup/cleanup logic in "list" -- which was doing far more than what every other test does.

Pulling once this passes.

PR #465, opened 2019-03-31
Add ttl to functions deployed in spec test

### LINKED ISSUE #463: All functions created in spec tests should have a ttl

arikmaor:
Fixes #463

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -91,7 +91,7 @@
   cleanup:
     - bn remove $FUNC_NAME
   steps:
-    -   in: bn create node8 $FUNC_NAME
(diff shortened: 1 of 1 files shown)


test/spec.yml
```diff
@@ -179,23 +179,9 @@
 - test: Test deploy with TTL (number)
   setup:
     - export FUNC_NAME=ttl_numeric$RANDOM$RANDOM
-    - bn create node8 $FUNC_NAME
+    - bn create --config.ttl 10 node8 $FUNC_NAME
```
arielshaqed: So... now we have all these tests that `create --config.<something>` works, but not that actually having _any_ config value in `binaris.yml` works.  (No biggie, but still a change.)

PR #466, opened 2019-03-31
Generate random function names for spec tests

### LINKED ISSUE #462: Need better naming for function created during spec test

arikmaor:
This continues the work in #464 and really fixes #462

### DIFF
--- a/test/spec.yml
+++ b/test/spec.yml
@@ -86,8 +86,6 @@
                   bn logs foo --since 9s*
 
 - test: Test login (good-path)
-  setup:
-    - export FUNC_NAME=login$RANDOM$RANDOM
   cleanup:
     - bn remove $FUNC_NAME
   steps:
@@ -150,7 +148,6 @@
 
 - test: Test deploy (good-path)
   setup:
-    - export FUNC_NAME=deploy$RANDOM$RANDOM
     - bn create --config.ttl '2 hours' node8 $FUNC_NAME
   cleanup:
     - bn remove $FUNC_NAME
@@ -164,7 +161,7 @@
 
 - test: Test public deploy (good-path)
   setup:
-    - export FUNC_NAME=public_deploy$RANDOM$RANDOM
+    - export FUNC_NAME=public_$FUNC_NAME
     - bn create --config.ttl '2 hours' node8 $FUNC_NAME
   cleanup:
     - bn remove $FUNC_NAME
@@ -178,7 +175,6 @@
 
 - test: Test deploy with TTL (number)
   setup:
-    - export FUNC_NAME=ttl_numeric$RANDOM$RANDOM
     - bn create --config.ttl 10 node8 $FUNC_NAME
     - bn deploy $FUNC_NAME
   steps:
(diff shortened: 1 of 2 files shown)


rylandg (commented):
@arielshaqed I'm tagging you because I want your insights specifically.

I just want to say that this is a pretty distinct departure from the original agreement about spec tests. We initially believed spec tests should be incredibly readable and directly reflect exact user input and expected output. In fact, the initial spec-tests design didn't use hardcoded names. I initially wrote the framework using random names and was requested by the team to instead use readable hand generated ones instead.

 I'm not necessarily opposed to this new direction, I just felt it was important to remember what we originally set out to do here. Maybe there is a happy middle ground, where "spec" tests retain their purity and readability and we have alternative specs that serve as rigorous e2e testing.

arielshaqed:
> @arielshaqed I'm tagging you because I want your insights specifically.

Also I'm a reviewer :-/

> I just want to say that this is a pretty distinct departure from the original agreement about spec tests. We initially believed spec tests should be incredibly readable and directly reflect exact user input and expected output. In fact, the initial spec-tests design didn't use hardcoded names. I initially wrote the framework using random names and was requested by the team to instead use readable hand generated ones instead.

Fixed names turn out to be a problem, hence in #464 I made them all meaningful + random suffix.  That addresses all your issues.

My goals: correctness, readability, ease of writing (in decreasing order of importance!).  IMO this PR retains correctness, makes readability worse and improves ease of writing.

> I'm not necessarily opposed to this new direction, I just felt it was important to remember what we originally set out to do here. Maybe there is a happy middle ground, where "spec" tests retain their purity and readability and we have alternative specs that serve as rigorous e2e testing.

I tend to agree; we did have a F2F discussion.  I dislike the magic property of `FUNC_NAME`, and I dislike that it also does not go far enough.  OTOH `$RANDOM$RANDOM` is awkward, relies too much on repeating the same pattern, and doesn't entirely work (it cannot be used in a `step.out`).

- We could add an interpolation syntax, e.g. `{{random:prefix:id}}` might expand to a random ID that starts with `prefix` and repeats whenever the same `id` is used.
- We could use Jinja2 to implement that interpolation syntax.
- We could give up and use Ansible or similar, probably with some plugins we add to wrap boilerplate.
- Etc.

michaeladda (dismissed):
@rylandg I agree this deviates from the initial plan for the CLI spec to be used as "product" truth.
In practice, we no longer use the spec as basis for product definition as we discuss that elsewhere, and use the spec as an internal testing vehicle. As a result, this change makes perfect sense to me.

rylandg:
Understood, makes sense.

michaeladda:
@arikmaor should we close this? Seems obsolete...

arikmaor:
I actually still think we should do it unless we plan on abandoning this repo which is only used for this tests right now AFAIK

arielshaqed:
On Sun, 24 Nov 2019, 23:38 Arik Maor <notifications@github.com> wrote:

> I actually still think we should do it unless we plan on abandoning this
> repo which is only used for this tests right now AFAIK
>
For my money, we should migrate to use `specshell`. Then we can write in
actual JavaScript, and we can generate whatever randomness we need, where
we need it, using well-known and easy to read libraries.

>

arielshaqed:
Not gonna happen, CLI for a closed project.

PR #470, opened 2019-04-08
Feature entrypoint in files

rylandg:
I know it was initially discussed that we wanted to use JavaScript template strings but Michael and I realized it was not the best fit. There were a few problems but an example is the difficulty with nesting template strings for things like our node8 example (which itself uses template strings). Instead we opted to rely on simple `sed` like mechanism available via the `replace` package.


bergundy:
I still prefer templates over this method.
Have you considered this package https://www.npmjs.com/package/string-format?

rylandg:
Michael and I decided that based on the value this brings it's not worth the time to explore a more full fledged solution. A specific issue with the template strings was inability to lint, issues with nested templating in our examples, etc.

arielshaqed:
> Michael and I decided that based on the value this brings it's not worth the time to explore a more full fledged solution. A specific issue with the template strings was inability to lint, issues with nested templating in our examples, etc.

I'm worried we're painting ourselves into a corner.  E.g. I doubt a templating system will ever be lintable.  But by making sure our _interim_ solution is lintable we are creating the unintended feature, and thus the expectation, that it *is* lintable (Hyrum's Law).

I think we should have _fewer_ features on the interim solution.  Having the template lintable by accident is not a good way to lint it.  Adding a spec test that runs `bn create && eslint` is better.

rylandg:
@arielshaqed @bergundy Michael gave his 👍 for templating. Will work on that patch.

rylandg:
Barring a test failure, things should be good now.

lib/functionTemplates/template.js
```diff
@@ -0,0 +1,28 @@
+const format = require('string-format');
```
arielshaqed: This requires doubling up braces, which are used in both the Python and the JavaScript templates!

lib/functionTemplates/template.js
```diff
@@ -0,0 +1,28 @@
...
+
+const node8Template =
+`'use strict'
+
+/**
+ * @type {{ import("./binaris").Handler }}
+ */
+exports.{entrypoint} = async (body, context) => {
+  const name = context.request.query.name || body.name || 'World';
+  return \`Hello \${{name}}!\`;
```
arielshaqed: [LTS](https://en.wikipedia.org/wiki/Leaning_toothpick_syndrome), and I don't mean "Long Term Support".  This line has 5 escape sequences using 3 different escape characters.

lib/create.js
```diff
@@ -36,19 +38,31 @@ const create = async function create(functionName, functionPath,
...
   };
+
   // replace the generic function name with the actual name
   YMLUtil.delFuncConf(binarisConf, templateName);
   YMLUtil.addFuncConf(binarisConf, finalName, funcConf);
   // now write out all the files that have been modified
-  await fse.copy(templatePath, path.join(functionPath), { overwrite: false, errorOnExist: true });
+  await copy(templatePath, functionPath, { overwrite: false, errorOnExist: true });
+
+  // currently custom entry points are not supported in Java8
```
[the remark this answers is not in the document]
bergundy: Good point

lib/functionTemplates/template.js
bergundy: This is good enough IMO, not ideal but good enough.

lib/functionTemplates/template.js
```diff
@@ -0,0 +1,28 @@
...
+};`;
+
+const pythonTemplate =
+`def {entrypoint}(body, req):
+    name = req.query.get('name') or body.get('name') or 'World'
```
bergundy: This template is wrong, the second command is the context, accessing `context.query` is deprecated.

bergundy: > This template is wrong, the second command is the context, accessing `context.query` is deprecated.

If you don't solve here, please open an issue.

PR #472, opened 2019-05-01
Explicitly support ipv4 in invoke

rylandg:
Ipv6 is causing issues for people.

### DIFF
--- a/sdk/auth.js
+++ b/sdk/auth.js
@@ -20,6 +20,7 @@ const verifyAPIKey = async function verifyAPIKey(apiKey) {
       headers: {
         'X-Binaris-Api-Key': apiKey,
       },
+      family: 4,
       json: true,
     });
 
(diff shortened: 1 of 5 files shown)


rylandg:
Sorry that this was half-assed, I was rushing to get it done. I’ll update
for deploy right away. I can’t post context because it’s customer
sensitive, but I’ll message you on slack.

michaeladda:
@rylandg any update?

sdk/auth.js
```diff
@@ -20,6 +20,7 @@ const verifyAPIKey = async function verifyAPIKey(apiKey) {
       headers: {
         'X-Binaris-Api-Key': apiKey,
       },
+      family: 4,
```
vogre: I think no reason to use raw request here,  possible to use `getValidatedBody` to get all the benefits from reuse?

rylandg:
It did pass but doesn't look like it from here. Merging.

PR #476, opened 2019-06-23
Remove support for redis triggers

michaeladda:


michaeladda:
```
git grep -i trigger
Jenkinsfile:                    description: 'whether to avoid triggering precious at the end',
Jenkinsfile:                            string(name: 'TRIGGER_BRANCH', value: BRANCH_NAME),
Jenkinsfile:            trigger_precious("${BRANCH_NAME}", "${JOB_NAME}");
```
scoobie is trigger ~happy~ empty
precious removal is only the after next step (next stop is binaris/spice#596 )

vogre:
Not called trigger, but a `REDIS_PASS` environment variable is getting passed for Redis tests, can be omitted if no longer needed.

michaeladda:
> Not called trigger, but a `REDIS_PASS` environment variable is getting passed for Redis tests, can be omitted if no longer needed.

Nice catch. Thanks!

PR #478, opened 2019-07-07
Remove obsolete feedback command

michaeladda:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "binaris",
-  "version": "7.0.1",
+  "version": "8.0.0,
```
arielshaqed: Gaaah!  No close quote.  (Breaks syntax highlighting in GitHub, but also probably fails compilation horribly...).
```suggestion
  "version": "8.0.0",
```

michaeladda:
Thanks for spotting. Fixed.
As for feedback, I suspect shift will rely on its web site.

PR #479, opened 2019-08-28
Deflake stats test

michaeladda:


michaeladda:
now it failed due to perf (the stats hack worked). waiting for the perf to pass consistently too before merging.

michaeladda:
removed the perf and merging since it stills improve flakiness

PR #485, opened 2019-10-30
Remove stats support

michaeladda:
Since this is a hidden command, and not user visible, a patch version bump is enough.
This is part of an effort to remove abacus.


michaeladda:
CI pass here: https://ci.binaris.io/blue/organizations/jenkins/gobinaris%2Fmarathon/detail/feature-no-stats/1/pipeline (with marathon changes too)

PR #481, opened 2019-09-19
Update test for node10

vogre:

PR #486, opened 2019-11-24
Update spec tests for node12

arikmaor:


test/jsApi.yml
```diff
@@ -161,5 +161,5 @@
              HTTP/1.1 231 unknown
              foo: 123
              bar: vaz
-             but:
```
vogre: Note: significant space.

vogre (approved):
Thanks!

Note #481 has some changes in the bad user code tests.

Approved assuming tests pass.

test/jsApi.yml
arikmaor: 10x!

